PR #90, opened 2017-04-11
Implement Pid functionality for the Services Query

hammol04:


BPScott:
We're still missing the repo and service tests for `findAllInNetworks`, but we can get those in a separate PR.

PR #91, opened 2017-04-25
Adding CacheItemPoolInterface to Service constructors

votemike:


Jayflux:
🎈 Happy First PR :champagne:

BjornTwachtmann:
Looks good to me :)

A few points that aren't strictly related to this code but I'm going to stick on here anyway because I'm an akward bastard...
a) We should make this a major version bump as it's a breaking change
b) We really need to start refering to this library by version in faucet/clifton/frontend. I think they're still pointing at dev-master.

BPScott:
@BjornTwachtmann: Agreed on A. I'm less convinced by B as that'll result in a lot of version number churn. The fatigue/overhead of having to do version bumps was annoying when dealing with translations, I'm a little wary of reintroducing that here.

BjornTwachtmann:
We can discuss it tomorrow I guess, feel free to merge for now anyway.

PR #92, opened 2017-04-26
5516 cache item pool interface

votemike:
Updated variable name

### DIFF
--- a/src/Service/LoggingServiceFactory.php
+++ b/src/Service/LoggingServiceFactory.php
@@ -30,11 +30,11 @@ class LoggingServiceFactory
     public function __construct(
         EntityManagerInterface $entityManager,
         MapperFactory $mapperFactory,
+        CacheItemPoolInterface $cache,
         LoggerInterface $logger,
-        Stopwatch $stopwatch,
-        CacheItemPoolInterface $cacheItemPoolInterface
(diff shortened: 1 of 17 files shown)


BjornTwachtmann:
Looks good. For future reference if you want to make changes to something that's already been merged I'd probably either
a) Just commit it straight to master if it's as simple as this and you're confident it will work
b) Create a separate branch and raise a PR for that if it's more complex

src/Service/LoggingServiceFactory.php
```diff
@@ -32,9 +32,9 @@ public function __construct(
         MapperFactory $mapperFactory,
         LoggerInterface $logger,
         Stopwatch $stopwatch,
-        CacheItemPoolInterface $cacheItemPoolInterface
+        CacheItemPoolInterface $cacheItemPool
```
BPScott: A little thing I missed last time -  The LoggingServiceFactory mimics the constructor of the ServiceFactory, but with two extra dependencies on the end. Thus I think adding the CacheItemPoolInterface after the MapperFactory rather than at the end of the constructor args might be handy to keep that symmetry.

votemike: Updated

PR #97, opened 2017-05-09
'Optimizing Imports'

votemike:
Ran Intelliji's Optomizing Imports in order to get rid of unused imports.

### DIFF
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -2,9 +2,9 @@
 
 namespace BBC\ProgrammesPagesService\Cache;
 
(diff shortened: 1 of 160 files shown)


BjornTwachtmann:
Bloody hell, that's a lot of changes. 

None of this _should_ cause any problems. It makes me a bit nervous in that it touches parts of code that are used very infrequently by Faucet, but can cause some nasty explosions if they go wrong.

It should be fine...

votemike:
I thought a tidy-up before tagging a version would be good.

PR #98, opened 2017-05-09
Change cache dependecy injection and implement cache class

BjornTwachtmann:


src/Cache/Cache.php
```diff
@@ -0,0 +1,121 @@
...
+    public function getItem(string $key): CacheItemInterface
+    {
+        $key = $this->standardiseKey($key);
+        if ($this->flushCacheItems) {
+            $this->cachePool->deleteItem($key);
+        }
+        return $this->cachePool->getItem($key);
+    }
+
+    public function setItem(CacheItemInterface $item, $ttl): bool
```
votemike: Is $ttl an int?  
Also, should there be a docblock to say that the ttl is in seconds? Or is that obvious enough?

src/Cache/Cache.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private function calculateTtl($ttl): int
+    {
+        if (is_numeric($ttl)) {
+            $ttl = (int) $ttl;
+        } elseif (is_string($ttl) && isset($this->cacheTimes[$ttl])) {
+            $ttl = $this->cacheTimes[$ttl];
+        } else {
+            throw new \InvalidArgumentException("TTL must be a number or Cache class constant");
```
votemike: `use InvalidArgumentException`

src/Cache/Cache.php
```diff
@@ -0,0 +1,121 @@
...
+        }
+        return $this->protectLifetimeFromStampede($ttl);
+    }
+
+    private function protectLifetimeFromStampede(int $ttl): int
```
votemike: I know what this does. But someone in 6 months may not, should this have a docblock?

src/Cache/Cache.php
BPScott: can be an int or string (sort, normal, etc, see the constants in the interface

src/Service/ProgrammesService.php
```diff
@@ -126,6 +126,26 @@ public function findByPid(Pid $pid, string $entityType = 'Programme'): ?Programm
         return $this->mapSingleEntity($dbEntity);
     }
 
+    /*
```
BPScott: This is !!! but it's gonna be removed once we move towards integrating these changes, so it's fine for the moment

BjornTwachtmann: It's there for reference. The fact that PHPCS hates it does not concern me...

BjornTwachtmann:
Comments added as per review. See if those make sense to you.

src/Cache/Cache.php
```diff
@@ -0,0 +1,160 @@
...
+    private $cachePool;
+
+    /** @var string */
+    private $prefix;
+
+    /** @var bool */
+    private $flushCacheItems = false;
+
+    private $defaultCacheTimes = [
+        CacheInterface::NONE => -1,
```
votemike: Did you find out what the correct value for this is?

src/Cache/Cache.php
```diff
@@ -0,0 +1,160 @@
...
+use Psr\Cache\CacheItemPoolInterface;
+use InvalidArgumentException;
+
+class Cache implements CacheInterface
+{
+    /** @var CacheItemPoolInterface */
+    private $cachePool;
+
+    /** @var string */
+    private $prefix;
```
votemike: What prefix is this? The environment?

src/Cache/Cache.php
```diff
@@ -0,0 +1,160 @@
...
+     * @param int|string $ttl
+     *   TTL in seconds, or a constant from CacheInterface
+     * @param callable $function
+     * @param array|null $arguments
+     * @return mixed
+     */
+    public function getOrSet(string $key, $ttl, callable $function, array $arguments = [])
+    {
+        $cacheItem = $this->getItem($key);
+        if ($cacheItem->isHit() && !$this->flushCacheItems) {
```
votemike: If `$this->flushCacheItems`, then the item is deleted as part of `$this->getItem($key)`, so do you need this test here too?

src/Cache/Cache.php
```diff
@@ -0,0 +1,160 @@
...
+        return $result;
+    }
+
+    /**
+     * Helps you to construct good cache keys by prodding you in the correct direction.
+     * Entirely optional but you are encouraged to use it.
+     *
+     * @param string $className
+     * @param string $functionName
+     * @param \string[] ...$uniqueValues
```
votemike: Does this need escaping?

src/Cache/Cache.php
BPScott: If you want to store something in the cache and have it instantly expire then -1. If you want to store something indefinitely then 0

src/Cache/Cache.php
BjornTwachtmann: yeah, you do, the cache item object is unaffected by the delete operation

src/Cache/Cache.php
BjornTwachtmann: A prefix, up to the calling application but generally the application version so that changes don't result in invalid objects being unserialised etc.

src/Cache/Cache.php
BjornTwachtmann: IntelliJ says it does. I believe everything intellij tells me implicitly.

votemike: All obey the Almighty Intellij.  
Almighty Intellij

src/Cache/Cache.php
votemike: Does that need some sort of docblock? ❤️ Docblocks 😒

src/Cache/Cache.php
votemike: When would you want something to expire instantly?

BjornTwachtmann: You probably wouldn't. I thought might as well leave it there for reference, though that may be stupid

tests/Service/LoggingServiceFactoryTest.php
```diff
@@ -47,7 +47,7 @@ public function testFactoryCallingNonExistantFunctionThrowsError()
         $service = new VersionsService(
             $this->createMock('BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\VersionRepository'),
             $this->createMock('BBC\ProgrammesPagesService\Mapper\ProgrammesDbToDomain\VersionMapper'),
-            new NullAdapter()
+            $this->createMock('BBC\ProgrammesPagesService\Cache\CacheInterface')
```
BjornTwachtmann: Agree in principle, we should probably do this in bulk rather than one at a time though.

PR #99, opened 2017-05-10
Additional funcionality

tatitati:
Not sure if this should be here, let me know if I'm wrong.

### DIFF
--- a/src/Service/BroadcastsService.php
+++ b/src/Service/BroadcastsService.php
@@ -53,7 +53,7 @@ class BroadcastsService extends AbstractService
         return $this->mapManyEntities($dbEntities);
     }
 
-    public function groupBroadcastsByPeriodOfDay($broadcasts, DateTimeImmutable $selectedDate = null)
+    public function groupBroadcastsByPeriodOfDay(array $broadcasts, DateTimeImmutable $selectedDate)
     {
         $intervalsDay = [
             'early' => [],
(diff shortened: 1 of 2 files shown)


src/Service/BroadcastsService.php
```diff
@@ -51,4 +52,71 @@ public function findByServiceAndDateRange(
...
+            'early' => [],
+            'morning' => [],
+            'afternoon' => [],
+            'evening' => [],
+            'late' => []
+        ];
+
+        $prior_broadcast = null;
+        foreach($broadcasts as $broadcast) {
+            // // If the end of the prior is earlier than the start of this broadcast
```
tatitati: I know we are not handling gasp broadcast for now in frontend, so not sure what to do regarding to this.

src/Service/BroadcastsService.php
```diff
@@ -51,4 +52,71 @@ public function findByServiceAndDateRange(
...
+    {
+        $dayStart = $selectedDate->setTime(0,0,0);
+        $dayEnd = $selectedDate->setTime(23,59,59);
+
+        $startBroadcast = $broadcast->getStartAt();
+        $startBroadcastHour = $startBroadcast->format('H');
+
+        switch ($startBroadcastHour)  {
+            case ($startBroadcast > $dayEnd && $startBroadcastHour < 6):
+                return 'late';
```
tatitati: Should I add these "late/evenint/..." as constants?

BjornTwachtmann:
This is mostly front end logic, I don't think it really belongs in here myself...

PR #100, opened 2017-05-11
Using cache for services

votemike:


BjornTwachtmann:
Wow, quick work. Couple of questions/comments...

1. We don't want to merge this until it's been load tested, so hold off on that
2. I think we might still need something in the cache class to handle a "don't cache this" case, or am I missing something?
3. Do we want the library to have different default cache lifetimes for different methods, i'm honestly not sure, maybe Ben has an opinion.

votemike:
I don't hang about....

1. Putting it live would load test it.
2.I haven't added that. This change is large enough and I felt it was a different piece of work.
3. @BPScott ?

BPScott:
1. I think what @BjornTwachtmann is after is a branch of Clifton that uses this, that we then loadtest on INT using the NullAdapter (i.e. PROGRAMMES-5529) and then if we're feeling particularly clever, the RedisAdapter. He's rather fastidious when it comes to perf regressions and hey we've got the tools to enable that streak so why not :D
2. What sort of thing are ya referring to? If we don't need it for this change I agree with @votemike, this PR is meaty enough already. It's nice to split "bulk changes, brainlessly scroll through" and "this is something we should think about" changes into separate PRs.
3. I reckon give everything a the standard 300 seconds to start with, then we can tweak the per-method defaults as required in new pr(s). Otherwise it'll be tough to see what deviates from the norm as part of this PR.

BPScott:
Bikeshedding! I really like the callback approach but there are few ways of writing it that all do the same thing but are formatted slightly differently. It'd be good to agree on a single formatting and apply it to all.

Big Q: Should we pass in an argument list or use `use`? Currently we pass in a final array of arguments, that are then passed to the function. Alternatively we could use `use` to make the function a closure and avoid this additional argument. Personally I like the closure idea, it saves an additional boilerplate list of arguments, and making this bunch of boilerplate less boilerplatey is good. It's kinda annoying with 2 items but in many cases we've got at least 3 args (when we have a `$limit` and `$page` for pagination).

Pure function:
```php
public function findByPidFull(Pid $pid, string $entityType = 'Programme', $ttl = CacheInterface::NORMAL): ?Programme
{
    $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $ttl, (string) $pid, $entityType);

    return $this->cache->getOrSet(
        $key,
        $ttl,
        function(Pid $pid, string $entityType) {
            $dbEntity = $this->repository->findByPidFull($pid, $entityType);

            return $this->mapSingleEntity($dbEntity);
        },
        [$pid, $entityType]
    );
}
```

Closure (using `use`):
```php
public function findByPidFull(Pid $pid, string $entityType = 'Programme', $ttl = CacheInterface::NORMAL): ?Programme
{
    $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $ttl, (string) $pid, $entityType);

    return $this->cache->getOrSet(
        $key,
        $ttl,
        function() use($pid, $entityType) {
            $dbEntity = $this->repository->findByPidFull($pid, $entityType);

            return $this->mapSingleEntity($dbEntity);
        }
    );
}
```

---

Alternatively @votemike has suggested using two functions - one private one that is the uncached query and one that the cache wrapping. This is nice as it forces a split between the cache bits and the data call, however I wonder if there is a net-win for this indirection. It gives us an option to make the raw query public to give people a chance to use that. But we end up duplicating argument lists which feels wasteful and creates more boilerplate, which I'd like to avoid.

```php
public function findByPidFull(Pid $pid, string $entityType = 'Programme', $ttl = CacheInterface::NORMAL): ?Programme
{
    $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $ttl, (string) $pid, $entityType);
    return $this->cache->getOrSet($key, $ttl, [$this, 'findByPidRaw'], [$pid, $entityType]);
}

private function findByPidFullRaw(Pid $pid, string $entityType = 'Programme'): ?Programme
{
    $dbEntity = $this->repository->findByPidFull($pid, $entityType);
    return $this->mapSingleEntity($dbEntity);
}
```

---

Any particular preferences?

TL;DR: This would be easier if PHP had built in annotations :<

BjornTwachtmann:
I'm honesty not sure whether having the function inline is more or less readable. The bigger function calls do start to look awkward when put inline, but duplicating every function is hardly ideal either.

On balance I'd probably go with the inline style, but either works really.

tatitati:
After having a look into both, I really like the use of "use", for me is very readable. I really like also closures, but I feel that they make sometimes more difficult testing/debugging. But anyway, I feel that both are good options, closures can create very compact code.

BPScott:
Personal reckons: The new functions for uncached items shall result in a lot of noise. I prefer the anonymous function approach.

I also prefer the `use` closure approach, I not needing that second argument list is a nice reduction in complexity.

Vote time! React to this post with:
* :+1: for anon functions with passing the arg list (i.e. Pure function example above)
* 😆 for anon functions using closures (i.e. the closure example above)
* 🎉 for two function approach.

/cc @BjornTwachtmann, @tatitati, @Tasiobg, @zoranikolic, @felipeparaujo, @votemike

BjornTwachtmann:
As stated before, I have no particularly strong opinions here, any of these are perfectly readable. I'd marginally favour the inline function with the use style, but if you can't be arsed to refactor (I wouldn't be), how it is currently seems perfectly fine to me.

BPScott:
5 people in favour of the use (with varying degrees of strength of preference) vs 1 for split methods, means closures win, sorry @votemike.

Starting as we mean to go on would be good, could you update this lot to use the closure style syntax and then this can be merged.

votemike:
Done

src/Cache/Cache.php
```diff
@@ -94,13 +94,13 @@ public function getOrSet(string $key, $ttl, callable $function, array $arguments
...
-     * @param \string[] ...$uniqueValues
+     * @param \null[]|\string[] ...$uniqueValues
      * @return string
      */
-    public function keyHelper(string $className, string $functionName, string ...$uniqueValues): string
+    public function keyHelper(string $className, string $functionName, ?string ...$uniqueValues): string
     {
         // Please help prevent cache namespace collisions by driving carefully
-        $uniqueValues = str_replace('.', '_', $uniqueValues);
+        $uniqueValues = str_replace('.', '_', array_filter($uniqueValues));
```
BPScott: Does this cause potential for collisions if there are multiple arguments that could be null?

Setting $uniqueValues to`['foo', 1, null]` or `['foo', null , 1]` would both result in the same key. Those empty values should be set to some placeholder value rather than stripped out entirely to avoid that collision.

src/Service/CollapsedBroadcastsService.php
```diff
@@ -102,145 +137,198 @@ public function findByCategoryAndStartAtDateRange(
...
     ): array {
         if (empty($allCategories)) {
             return [];
         }
 
-        $categoriesAncestryIds = [];
-        foreach ($allCategories as $category) {
-            $categoriesAncestryIds[] = $category->getDbAncestryIds();
-        }
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, md5(serialize($allCategories)), $from->getTimestamp(), $to->getTimestamp(), $ttl);
```
BPScott: Build an array of the category Ids to use in the key, rather than `md5(serialize($allCategories))`.

In the segments service you do this imploding on '-'.

src/Service/CategoriesService.php
```diff
@@ -18,39 +18,79 @@ public function __construct(
...
+            }
+        );
     }
 
-    public function findGenreByUrlKeyAncestry(array $urlHierarchy): ?Genre
+    public function findGenreByUrlKeyAncestry(array $urlHierarchy, $ttl = CacheInterface::NORMAL): ?Genre
     {
-        $genre = $this->repository->findByUrlKeyAncestryAndType($urlHierarchy, 'genre');
-        return $this->mapSingleEntity($genre);
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode($urlHierarchy), $ttl);
```
BPScott: implode using '-' as you do that everywhere else (or update the other places to not use a hyphen)

Tasiobg:
Good work @votemike 

I think this requires a good testing, it's easy to miss something with that many changes.

I was wondering what is the plus point of having the $ttl as a part of the $key.
I also wonder If different apps can use the "programmes-pages-service" library which one of the next case are we in (can be a non listed one)?
- Case 1: All keys are shared between any app using "programmes-pages-service", keys should share same namespace to avoid unnecesary keys duplications and redis writes. In this case the $ttl should be set inside the service function and shouldn't change
- Case 2: Any app using "programmes-pages-service" has their own namespace so keys are unique to that app, different apps can have different $ttl
- Case 3: Some apps share the same namespace while others apps have their own namespace.

@BjornTwachtmann @BPScott what are your thoughts about the text above? which case are we in?

Tasiobg:
Other thing I would like to mention, I miss the PR for Cache.php, I think we could implement the "don't cache this" mechanims by updating the "getOrSet" function so when we set CacheInterface::NONE the function will not cache anything (open to more investigation what to do with getting items from cache if CacheInterface::NONE is set). Somenthing else very minor and only personal opinion, I suggest to rename the $flushCacheItems and avoid the "flush" part because it could be confused with the redis flushall operation. All this could be on a different PR, this one is already quite big

src/Service/NetworksService.php
```diff
@@ -17,23 +17,40 @@ public function __construct(
...
-        int $page = self::DEFAULT_PAGE
+        int $page = self::DEFAULT_PAGE,
+        $ttl = CacheInterface::NORMAL
     ): array {
-        $dbEntities = $this->repository->findPublishedNetworksByType(
-            $types,
-            $limit,
-            $this->getOffset($limit, $page)
-        );
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode($types), $limit, $page, $ttl);
```
BPScott: implode using '-' as you do that everywhere else (or update the other places to not use a hyphen)

BPScott:
> I was wondering what is the plus point of having the $ttl as a part of the $key.

It avoids race conditions. The TTL deals with the age of the new thing that can go in the cache rather than checking for the age of the anything a subsequent call finds in the cache. Consider two pages that call the same service function but with different cache times, e.g.
* `$programmesService->findByPid(new Pid('b0000001'), 36000)` is called at 3pm. It caches for 10 hours.
* `$programmesService->findByPid(new Pid('b0000001'), 300)` is called at 7pm. It caches for 5 minutes.

If you don't include the cache TTL as part of the key then that second result shall return the cached result that was put into the cache at 3pm - fours hours ago, despite it asking for a result that it expects to be at most 5 minutes old.

src/Cache/Cache.php
votemike: Is replacing null with a blank string acceptable?

src/Service/NetworksService.php
votemike: It was my intention that each part of the key was split by a hyphen. If I imploded the types by hyphen too, it would not make the key as clear if we were to inspect it.

src/Service/CategoriesService.php
votemike: It was my intention that each part of the key was split by a hyphen. If I imploded the urlHierarchy by hyphen too, it would not make the key as clear if we were to inspect it.

src/Service/CollapsedBroadcastsService.php
votemike: That was a mistake. It was my intention that each part of the key was split by a hyphen. If I imploded the categoryIds by hyphen too, it would not make the key as clear if we were to inspect it. I would prefer that I update the method in the programmesService. What do you think.

votemike:
@Tasiobg Do you mean unit tests or manual testing? I have updated the unit tests. I'm not sure what other tests I could write to test the caching.  
Or if you mean manual testing, I hope the testers will do that.

votemike:
@Tasiobg @BPScott Does one of you want to open a ticket to implement `CacheInterface::NONE`?

Tasiobg:
@votemike I mean tested in general, I was a bit worried because of the amount of changes and I wanted to see if someone mention anything, not sure how we test things but looks like we have a lot of unitest wich is nice

src/Cache/Cache.php
BPScott: Yep, I think that'd result in `foo_1` and `foo__1` in the above two examples. It's a bit opaque though, is there some other magic character that we could use instead? This list of valid cache keys is pretty small though :(

Tasiobg:
@BPScott I can be wrong because I'm not full aware of all posible cases on this application but I think we should try to use only one $ttl per key, this will made code more consistent and easy to maintain.
What is the reason to have 2 $ttl for the same key and one lasting 5 minutes and other lasting hours? won't make more sense to use allways the shorter one which is more recent?

The way I'm use to implement caching keys and TTL is by enums or conf files.
This is an example of how to centralize all the key and TTL on one enum class, bear on mind that there is a "%s" which is meant to be replace with the variable part of the key when setting or getting the key.

```
class CacheEnums
{

    public static $AtozTitlesKeys = [
        'findTleosByFirstLetter' => 'BBC.ProgrammesPagesService.Service.findTleosByFirstLetter.%s',
        'countTleosByFirstLetter' => 'BBC.ProgrammesPagesService.Service.countTleosByFirstLetter.%s'
        ];

    public static $AtozTitlesTtl = [
        'findTleosByFirstLetter' => CacheInterface::NORMAL,
        'countTleosByFirstLetter' => CacheInterface::NORMAL
    ];
}
```

With the example above you can have all application keys and TTL in one centralized class, this helps to understand what is cached and for how long at a glance.
In this case you can't autogenerate the key names, only the variable part can be append to the already existing enum.
I think I like this approach a bit more but just personal opinion.

@votemike @BjornTwachtmann  feel free to share your opinion as well

src/Service/ProgrammesService.php
```diff
@@ -38,9 +38,17 @@ public function __construct(
         parent::__construct($repository, $mapper, $cache);
     }
 
-    public function countAllTleosByCategory(Category $category): int
+    public function countAllTleosByCategory(Category $category, $ttl = CacheInterface::NORMAL): int
     {
-        return $this->repository->countTleosByCategory($category->getDbAncestryIds(), false);
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode('-', $category->getDbAncestryIds()), $ttl);
```
BPScott: Use $category-> getDbId(), rather than the list of all its parent ancestry. You do this a few times in this file.

src/Service/NetworksService.php
BPScott: If you don't implode with a delimiter then you end up with cache collisions between [3,2] and [32] as they would both generate the key part '32'.

src/Service/CategoriesService.php
BPScott: If you don't implode with a delimiter then you end up with cache collisions between [3,2] and [32] as they would both generate the key part '32'.

src/Service/CollapsedBroadcastsService.php
BPScott: If you don't implode with a delimiter then you end up with cache collisions between [3,2] and [32] as they would both generate the key part '32'.

I think the imploding with a hypen is a more correct behaviour as it avoids those collisions.

BPScott:
@Tasiobg, Admitedly it was a contrived example. I agree that we should strive to only use the one TTL the vast majority of the time. However there may be occasional cases where we need a different lifetime for a particular request. This is why we offer the ability to override the default TTL as an the final argument to a service. The vast majority of the time people consuming this lib won't need to specify that argument and a sensible default (CacheInterface::NORMAL) shall be used.

src/Service/ContributionsService.php
```diff
@@ -22,68 +22,103 @@ public function __construct(
...
         ?int $limit = self::DEFAULT_LIMIT,
-        int $page = self::DEFAULT_PAGE
+        int $page = self::DEFAULT_PAGE,
+        $ttl = CacheInterface::NORMAL
     ): array {
-        $segmentIds = array_map(function ($s) {
+        $segmentIds = array_map(function (Segment $s) {
             return $s->getDbId();
         }, $segments);
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode('-', $segmentIds), $limit, $page, $ttl);
```
BPScott: Implode using | as in the other places?

PR #101, opened 2017-05-15
Add Broadcast->isOnAir and Service->isActive

BPScott:
Add clarifying comments on if start/end times for services and
broadcasts are inclusive or exclusive (starts are inclusive,
ends are exclusive)

### DIFF
--- a/src/Domain/Entity/Broadcast.php
+++ b/src/Domain/Entity/Broadcast.php
@@ -101,11 +101,23 @@ class Broadcast
         return $this->service;
     }
 
+    /**
+     * When the broadcast started. This is inclusive, this is the very first
+     * moment of the broadcast.
+     * Given two broadcasts, the first will end at at 06:30:00 and the second
+     * will start at 06:30:00.
+     */
     public function getStartAt(): DateTimeImmutable
     {
         return $this->startAt;
     }
 
+    /**
+     * When the broadcast ended. This is exclusive, this is the very first
+     * moment that the broadcast is no longer on.
+     * Given two broadcasts, the first will end at at 06:30:00 and the second
+     * will start at 06:30:00.
+     */
     public function getEndAt(): DateTimeImmutable
     {
         return $this->endAt;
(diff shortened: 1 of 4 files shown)


votemike:
How does this deal with the switch between BST and GMT? E.g. if a broadcast starts at 01:45 and ends at 01:15?

src/Domain/Entity/Service.php
```diff
@@ -125,4 +140,17 @@ public function getLiveStreamUrl(): ?string
...
+
+    /**
+     * Returns true if the service is active on a given date.
+     * If a service has no start date then we assume it was active since the
+     * dawn of time.
+     * If a service has no end date then we assume it will be active until the
+     * end of time.
+     * Service starts are inclusive and ends are exclusive.
+     */
+    public function isActive(DateTimeImmutable $date): bool
```
votemike: `isActiveOn`?

BjornTwachtmann:
Logic looks sound, I hope there aren't as many nasty lurking timezone issues in DateTime as in our old code, there shouldn't be really. I agree isActiveOn makes more sense if a date is being passed in.

src/Domain/Entity/Broadcast.php
```diff
@@ -125,4 +137,15 @@ public function isRepeat(): bool
...
         return $this->isRepeat;
     }
+
+    /**
+     * Returns true if the broadcast is on air at a given date.
+     * Broadcast starts are inclusive and ends are exclusive.
+     * Given two broadcasts, the first will end at at 06:30:00 and the second
+     * will start at 06:30:00.
+     */
+    public function isOnAir(DateTimeImmutable $date): bool
```
votemike: I know it's fussy as it's only a variable name used internally, but this isn't a date. It's a date/time. So should the variable actually be `$datetime`?  
If this was in a single project, I wouldn't be this fussy, but it's in a composer package that may be used by a number of people and so correctness/readabilty seems important.

BPScott:
All times should be in the UTC timezone, with a TZ signifying it as such. We shall deal with the conversion to UK time (GMT or BST) at render time.

Shall update to `isActiveOn`. Should I change `isOnAir` to `isOnAirOn` for the same reasons? That repetition  of 'on' looks odd, 'isBroadcastingOn`?

Hopefully the move to using`DateTimeImmutable` shall help avoid much of the weird around mutable date instances.

src/Domain/Entity/Broadcast.php
BPScott: Good spot. Technically correct is the best kind of correct.

BPScott:
Have updated the functions to be `isActiveAt` and `isOnAirAt` and am passing in `$dateTime` to both of them

PR #103, opened 2017-05-25
Adding support for BST/UTC

votemike:


src/Data/ProgrammesDb/Type/UtcDateTimeType.php
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        $utcValue = $value->setTimezone(new DateTimeZone('UTC'));
+
+        return parent::convertToDatabaseValue($utcValue, $platform);
+    }
+
+    public function convertToPHPValue($value, AbstractPlatform $platform): ?DateTimeInterface
+    {
+        if (null === $value || $value instanceof DateTimeInterface) {
```
votemike: Are we happy with the check for `DateTimeInterface`? Or should we check `DateTimeImmtable` so that only `null` or `DateTimeImmutable`s are returned?

README.md
```diff
@@ -102,6 +102,24 @@ You can repeat the `pps.programmes_service` definition, changing the
...
+```yaml
+doctrine:
+    dbal:
+        types:
+            datetime: BBC\ProgrammesPagesService\Data\ProgrammesDb\Type\UTCDateTimeType
+            datetimetz: BBC\ProgrammesPagesService\Data\ProgrammesDb\Type\UTCDateTimeType
+```
+
+You can also force Twig to format date's using a timezone of your choice by adding the following to your config.
+```yaml
```
BjornTwachtmann: I wouldn't document the twig stuff here, it's not releavant to pages service (minor nitpick)

src/Data/ProgrammesDb/Type/UtcDateTimeType.php
```diff
@@ -0,0 +1,50 @@
...
+{
+    static private $utc;
+
+    public function convertToDatabaseValue($value, AbstractPlatform $platform): ?string
+    {
+        if (!$value instanceof DateTimeInterface) {
+            throw new InvalidArgumentException('$value must implement DateTimeInterface');
+        }
+
+        $utcValue = $value->setTimezone(new DateTimeZone('UTC'));
```
BjornTwachtmann: if I pass in a \DateTime (not immutable) object, won't this modify it? Or am I being thick?

src/Data/ProgrammesDb/Type/UtcDateTimeType.php
BjornTwachtmann: I think that's fine

BjornTwachtmann:
One minor point, otherwise, perfect

src/Data/ProgrammesDb/Type/UtcDateTimeType.php
votemike: I guess. Yeah. How would you avoid that? Only allow `DateTimeImmutable`s to be passed in?

BPScott:
Adding new Doctrine types gives me the fear. I'd like to learn a bit more about whats going on, lets chat tomorrow when we're both in.

src/Data/ProgrammesDb/Type/UtcDateTimeType.php
BjornTwachtmann: Duplicate the DateTime if it's not immutable I guess. Just try it really.

BPScott: Changing this to expect DateTimeImmutable only will wreck merry havoc as Doctrine expects to send and return DateTime (not Immutable). Making that inconsistent with this that default behaviour will cause weirdness.

BjornTwachtmann: We pass in Immutable all the time. It's not required though...

PR #104, opened 2017-05-30
Datetime fix

votemike:
It seems `getDomainModel()` was expecting a date string in the array, but now a DateTimeImmutable is being passed in.

### DIFF
--- a/src/Mapper/ProgrammesDbToDomain/ProgrammeMapper.php
+++ b/src/Mapper/ProgrammesDbToDomain/ProgrammeMapper.php
@@ -13,7 +13,6 @@ use BBC\ProgrammesPagesService\Domain\Entity\Unfetched\UnfetchedMasterBrand;
 use BBC\ProgrammesPagesService\Domain\Entity\Unfetched\UnfetchedProgramme;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
-use DateTimeImmutable;
 use InvalidArgumentException;
 
 class ProgrammeMapper extends AbstractMapper
@@ -88,7 +87,7 @@ class ProgrammeMapper extends AbstractMapper
             $this->getMasterBrandModel($dbProgramme),
             $this->getCategoriesModels('genre', $dbProgramme, 'categories'),
             $this->getCategoriesModels('format', $dbProgramme, 'categories'),
-            ($dbProgramme['firstBroadcastDate'] ? DateTimeImmutable::createFromMutable($dbProgramme['firstBroadcastDate']) : null),
+            $this->castDateTime($dbProgramme['firstBroadcastDate']),
             $dbProgramme['expectedChildCount']
         );
     }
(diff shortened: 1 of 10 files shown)


src/Mapper/ProgrammesDbToDomain/ProgrammeMapper.php
```diff
@@ -88,7 +87,7 @@ private function getBrandModel(array $dbProgramme): Brand
             $this->getMasterBrandModel($dbProgramme),
             $this->getCategoriesModels('genre', $dbProgramme, 'categories'),
             $this->getCategoriesModels('format', $dbProgramme, 'categories'),
-            ($dbProgramme['firstBroadcastDate'] ? DateTimeImmutable::createFromMutable($dbProgramme['firstBroadcastDate']) : null),
+            $dbProgramme['firstBroadcastDate'],
```
votemike: Is this correct? I assume if `$dbProgramme['firstBroadcastDate']` isn't truthy, it's null, so no need to specifically set it as null?

src/Mapper/ProgrammesDbToDomain/ServiceMapper.php
```diff
@@ -33,8 +32,8 @@ public function getDomainModel(array $dbService): Service
                 $dbService['shortName'],
                 $dbService['urlKey'],
                 $this->getNetworkModel($dbService),
-                ($dbService['startDate'] ? DateTimeImmutable::createFromMutable($dbService['startDate']) : null),
-                ($dbService['endDate'] ? DateTimeImmutable::createFromMutable($dbService['endDate']) : null),
+                $dbService['startDate'],
```
votemike: Same question as file above.

src/Mapper/ProgrammesDbToDomain/ProgrammeMapper.php
BPScott: yep it'll either be a DateTimeImmutable or null.

src/Mapper/ProgrammesDbToDomain/ServiceMapper.php
BPScott: yep this is fine, it'll either be a DateTimeImmutable or null

tests/Mapper/ProgrammesDbToDomain/CollapsedBroadcastMapperTest.php
```diff
@@ -53,8 +53,8 @@ public function testGetDomainModel()
 
         $dbEntityArray = [
             'id'               => 1,
-            'startAt'          => new DateTime('2015-01-03T00:00:00'),
-            'endAt'            => new DateTime('2015-01-03T01:00:00'),
+            'startAt'          => new DateTimeImmutable('2015-01-03T00:00:00', new DateTimeZone('UTC')),
```
BPScott: In other tests we use zulu time `'2015-01-03T00:00:00Z'` to mark the timezone as UTC, can we do that here too, it might be sohrter than passing in a new DateTimeZone?

votemike: I wondered what the Z was for...  
Is Zulu forever UTC? It has no daylight savings?

BPScott: Yep z is always UTC, no daylight savings. (https://en.wikipedia.org/wiki/Coordinated_Universal_Time#Time_zones)

PR #105, opened 2017-05-31
Fetch Application Time in Europe/London time

votemike:


BjornTwachtmann:
I don't think we can have the default TZ as europe/london for anything that's not using the Doctrine DateTime type you wrote, shit will seriously break.

PR #108, opened 2017-06-01
Adding Chronos

votemike:


README.md
```diff
@@ -104,7 +104,9 @@ save you having to request it through the ServiceFactory every time.
 
 #### Timezone support
 By default, Doctrine2 assumes all DateTimes are in UTC.  
-You can force all dates going in to the DB to be converted to UTC and all DateTimes coming out to have UTC set by adding the following configuration to your YAML files.
+You can force all dates going in to the DB to be converted to UTC and all DateTimes coming out to have UTC set by adding the following configuration to your YAML files.  
+If using this configuration, DateTimes passed in to the service must be a [Chronos](https://book.cakephp.org/3.0/en/chronos.html) object.
```
BPScott: Do we need to care about the type that goes into the DB? We're only changing the output of what comes out (what was a DateTimeImmutable, is now a Cronos object) It seems like we'd still be able to accept any instance of DateTimeInterface, as before.

votemike: That makes it a pain to check if the interface is mutable or not. We'd need to clone it if it was mutable.

BPScott: Turns out we needed that for fixturing and created at bits, so this readme line can be removed.

src/Domain/ApplicationTime.php
```diff
@@ -69,6 +70,10 @@ public static function setTime(int $appTime = null)
     {
         static::$appTime = DateTimeImmutable::createFromFormat('U', $appTime ?? time())
             ->setTimezone(new DateTimeZone('UTC'));
+
+        if (class_exists(Chronos::class)) {
```
BPScott: Should we do this for the [other Chronos classes too](https://book.cakephp.org/3.0/en/chronos.html#testing-aids)? I can't see us ever using the Mutable classes, but `Date` feels like it'll come in handy in the future

votemike: `setTime` accepts a timestamp. I'm not sure how you would use it to set _just_ a date. I can also set the Chronos `Date` if you like.

BPScott: fair enough, let's cross that bridge when we get to it.

BjornTwachtmann:
looks good to me

PR #109, opened 2017-06-01
Add isOnAir method to broadcast domain object

BjornTwachtmann:


BPScott:
Is it worth adding this when we can call $broadcast->isOnAirAt(ApplicationTime::getTime())? in whatever calls this. I think it is worth trying to not let ApplicationTime pervade the whole application.

PR #110, opened 2017-06-03
Cast time() to a string for createFromFormat

djmarland:
`DateTimeImmutable::createFromFormat` expects a string as the second parameter. `time()` returns an int.
While ~~stealing your code~~ using your code as a great reference I've been setting strict types everywhere, so this one bit me.

Just a little bug you may encounter in future if you come to add strict types to this file.


BPScott:
Thanks @djmarland.

PR #113, opened 2017-06-06
[isite] domain-entities + mappers

tatitati:
Included in this PR:

1. We assume that we have data (options) in our new ref_options table in json format. 
2. We want to process these options and update our core_entities/network with the resulting-final options for each entity by studying the programmes hierarchy. This means that we have to update our entities to include the new attr "options" (in programmes and network)
3. This processing will be done for the  NetworkMapper, ProgrammesMaper. They analyse and assign the final options to the entities.


src/Domain/Entity/Unfetched/UnfetchedNetwork.php
```diff
@@ -12,7 +13,8 @@ public function __construct()
         parent::__construct(
             new NullNid(),
             '',
-            new UnfetchedImage()
+            new UnfetchedImage(),
+            new Options()
```
BPScott: Should be UnfetchedOptions, as we don't know what the options are, rather than "we know them any they're empty"

src/Domain/Entity/Unfetched/UnfetchedProgramme.php
```diff
@@ -22,7 +23,8 @@ public function __construct()
             false,
             false,
             false,
-            0
+            0,
+            new Options()
```
BPScott: Should be UnfetchedOptions, as we don't know what the options are, rather than "we know them any they're empty"

src/Domain/Entity/Unfetched/UnfetchedProgrammeItem.php
```diff
@@ -25,7 +26,8 @@ public function __construct()
             false,
             0,
             MediaTypeEnum::UNKNOWN,
-            0
+            0,
+            new Options()
```
BPScott: Should be UnfetchedOptions, as we don't know what the options are, rather than "we know them any they're empty"

src/Mapper/ProgrammesDbToDomain/MasterBrandMapper.php
```diff
@@ -5,6 +5,7 @@
 use BBC\ProgrammesPagesService\Domain\Entity\Image;
 use BBC\ProgrammesPagesService\Domain\Entity\MasterBrand;
 use BBC\ProgrammesPagesService\Domain\Entity\Network;
+use BBC\ProgrammesPagesService\Domain\Entity\Version;
```
BPScott: ?

tests/Data/ProgrammesDb/Entity/RefOptionsTest.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        new RefOptions(
+            'id',
+            'wrongwrongwrong'
+        );
+    }
+
+    private function mockCoreEntity()
+    {
+        return $this->createMock('BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\CoreEntity');
```
votemike: If you do `return $this->createMock(CoreEntity::class);` we can click in to the class with our IDEs. It also helps if we decide to rename the class in the future.

src/Domain/Entity/Network.php
```diff
@@ -173,4 +178,14 @@ public function isRadio(): bool
     {
         return ($this->getMedium() === NetworkMediumEnum::RADIO);
     }
+
+    public function getOptions(): Options
```
votemike: Is it possible that these are unfetched options?

BPScott: Options shall be a stored as a json column on the network table, it shall always be returned. It might be an empty, in which case an empty Options object shall be returned

votemike: Then why is there a `UnfetchedOptions` class as part of this PR?

tatitati: Options is the superclass of UnfetchedOptions. So the return type (generic) is Options as can be both. In this case UnfetchedOptions has only an empty array of options.

This said is the responsibility ProgrammeMapper to figure out if the whole hierarchy has been fetched. Otherwise the programme will have an UnfetchedOptions.

In the other side NetworkMapper doesn't need to check the hierarchy, as they are in the top hierarchy. So they set always Options.

Sometimes is also possible to have (before this new feature) to get an UnfetchedNetwork, UnfetchedEpisode, etc. For all of these Unfetched entities we set also UnfetchedOptions.

tests/Mapper/ProgrammesDbToDomain/MasterBrandMapperTest.php
```diff
@@ -2,11 +2,12 @@
 
 namespace Tests\BBC\ProgrammesPagesService\Mapper\ProgrammesDbToDomain;
 
-use BBC\ProgrammesPagesService\Domain\Entity\MasterBrand;
```
BPScott: The MasterBrandMapper has not changed, why do the tests need to be updated?

tests/Mapper/ProgrammesDbToDomain/OptionsMapperTest.php
```diff
@@ -0,0 +1,118 @@
...
+        $expectedEntity = new Options($expectedOptions);
+
+        $mapper = $this->getMapper();
+
+
+        $arrayOfAncestors = [$parentOptions, $grandparentOptions];
+
+        $this->assertEquals(
+            $expectedEntity,
+            $mapper->getDomainModel($options, ...$arrayOfAncestors)
```
BPScott: No need for the $arrayOfAncestors and using the spread operator, you could do:

`$mapper->getDomainModel($options, $parentOptions, $grandparentOptions)`

tests/Mapper/ProgrammesDbToDomain/ProgrammeMapperOptionsMappingTest.php
```diff
@@ -0,0 +1,226 @@
...
+            'id' => 1,
+            'mid' => 'bbc_one',
+            'network' => [
+                'id' => 1,
+                'options' => $networkOptions,
+            ],
+        ];
+
+        $expectedMasterBrandDomainEntity = $this->createMock(
+            'BBC\ProgrammesPagesService\Domain\Entity\MasterBrand'
```
BPScott: Use `MasterBrand::class`

tests/Mapper/ProgrammesDbToDomain/ProgrammeMapperOptionsMappingTest.php
```diff
@@ -0,0 +1,226 @@
...
+        );
+
+        $this->mockMasterBrandMapper->expects($this->once())
+            ->method('getDomainModel')
+            ->with($masterBrandDbEntity)
+            ->willReturn($expectedMasterBrandDomainEntity);
+
+        $expectedOptionsEntity1 = $this->createMock(
+            'BBC\ProgrammesPagesService\Domain\Entity\Options'
```
BPScott: use `Options::class`

tests/Mapper/ProgrammesDbToDomain/ProgrammeMapperOptionsMappingTest.php
```diff
@@ -0,0 +1,226 @@
...
+        );
+
+        $expectedOptionsEntity2 = $this->createMock(
+            'BBC\ProgrammesPagesService\Domain\Entity\Options'
```
BPScott: use `Options::class`

tests/Mapper/ProgrammesDbToDomain/ProgrammeMapperTest.php
```diff
@@ -17,6 +18,13 @@ class ProgrammeMapperTest extends BaseProgrammeMapperTestCase
 {
     public function testGetDomainModelBrand()
     {
+        $options = ['programmeOptions'];
```
BPScott: This test file doesn't test the mappers of any of the other models that require mappers (e.g.Image, Masterbrand) so there isn't a need to do that for options either.

In the tests in this file assume the options are blank - the value of `[]` and don't bother writing any expectations on the mockOptionsMapper. 

Testing that the Programme passes things through to the OptionMapper is done in the ProgrammeMapperOptionsMappingTest

BPScott:
Main source logic looks good, but there's a few little things in the tests that can be simplified as you're repeating yourself a little

PR #114, opened 2017-06-07
Add a utility methods to get valid values from enums

BPScott:
This saves us having to define the list of valid values in multiple
places

### DIFF
--- a/src/Data/ProgrammesDb/Entity/Network.php
+++ b/src/Data/ProgrammesDb/Entity/Network.php
@@ -192,12 +192,10 @@ class Network
 
     public function setMedium(string $medium): void
     {
-        if (!in_array($medium, [NetworkMediumEnum::RADIO, NetworkMediumEnum::TV, NetworkMediumEnum::UNKNOWN])) {
+        if (!in_array($medium, NetworkMediumEnum::validValues())) {
             throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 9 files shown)


votemike:
These all are used within in_array. Would it be better to pass the value in to the method and let the ENUM do the in_array check?

votemike:
Should all the ENUM extend or implement something with the method `validValues`?

BPScott:
> These all are used within in_array. Would it be better to pass the value in to the method and let the ENUM do the in_array check?

@votemike We'd still need to build the list of valid values that the enum has for the exception message.

BjornTwachtmann:
Seems reasonable. 

Some of the faucet stuff uses some reflection dark magic around enums to enable you to do things like have them check their list of constants for valid values without having to list them, instantiate just using the constant name, etc. I'm not entirely sold on using reflection in user facing code though, this seems less problematic.

votemike:
I didn't mean for you to replace `validValues()` but to go one step further and add an extra method to do that check.  
I'm easy either way.

Tasiobg:
I think having a BaseEnum class and extending it would make sense, at least this is what I have done in past projects and works fine. Example useful functions:
- getConstants() 
- getNameByValue()
- getValueByName()
- isValidName()
- isValidValue()

Related question on stackoverflow https://stackoverflow.com/questions/254514/php-and-enumerations

PR #115, opened 2017-06-08
Add BroadcastGap domain object

BPScott:
This is object similar to Broadcast, that can be used for handling
gaps in service transmission schedules

### DIFF
--- a/tests/Domain/Entity/BroadcastGapTest.php
+++ b/tests/Domain/Entity/BroadcastGapTest.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Tests\BBC\ProgrammesPagesService\Domain\Entity;
+
+use BBC\ProgrammesPagesService\Domain\Entity\BroadcastGap;
+use BBC\ProgrammesPagesService\Domain\Entity\Service;
+use BBC\ProgrammesPagesService\Domain\Entity\Unfetched\UnfetchedService;
+use DateTimeImmutable;
+use PHPUnit\Framework\TestCase;
+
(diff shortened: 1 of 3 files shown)


tests/Domain/Entity/BroadcastGapTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+use BBC\ProgrammesPagesService\Domain\Entity\BroadcastGap;
+use DateTimeImmutable;
+use PHPUnit\Framework\TestCase;
+
+class BroadcastGapTest extends TestCase
+{
+    public function testConstructorRequiredArgs()
+    {
+        $service = $this->createMock('BBC\ProgrammesPagesService\Domain\Entity\Service');
```
votemike: `$service = $this->createMock(Service::class);`?

tests/Domain/Entity/BroadcastGapTest.php
```diff
@@ -0,0 +1,50 @@
...
+        $startAt = new DateTimeImmutable('2015-01-01 06:00:00');
```
votemike: Should these have `Z` on the end?

BPScott: Doesn't really matter, nothing else in the Domain tests bother

tests/Domain/Entity/BroadcastGapTest.php
BPScott: None of the other Domain tests use this still, but they probably should.

PR #116, opened 2017-06-09
[isite] Update schema in ref_options table

tatitati:
To test follow next steps.

Faucet composer.json:
programmes-pages-service pointing this branch: `dev-update-db`

Then in faucet:
`php app/console doctrine:schema:update --dump-sql`

If you like the sql queries you can run them:
`php app/console doctrine:schema:update --force`

### DIFF
--- a/src/Data/ProgrammesDb/Entity/RefOptions.php
+++ b/src/Data/ProgrammesDb/Entity/RefOptions.php
@@ -2,14 +2,13 @@
 
 namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
 
-use DateTime;
 use Doctrine\ORM\Mapping as ORM;
-use Gedmo\Exception\InvalidArgumentException;
 use Gedmo\Timestampable\Traits\TimestampableEntity;
+use InvalidArgumentException;
 
 /**
  * @ORM\Entity()
- * @ORM\Table(indexes={@ORM\Index(name="ref_options_idx", columns={"guid", "entity"})})
+ * @ORM\Table(indexes={@ORM\Index(name="entity_id_idx", columns={"entity_id"})})
  */
 class RefOptions
 {
@@ -30,57 +29,55 @@ class RefOptions
     /**
      * @var string
      *
-     * @ORM\Column(type="string", length=50, nullable=false)
+     * @ORM\Column(type="string", length=36, unique=true, nullable=false)
      */
     private $guid;
 
     /**
      * @var string
      *
-     * @ORM\Column(type="string", length=50, nullable=false)
+     * @ORM\Column(type="string", length=35, nullable=false)
      */
     private $projectId;
 
     /**
-     * @var string
+     * @var integer
      *
-     * @ORM\Column(type="string", nullable=true)
+     * @ORM\Column(type="integer", nullable=false)
      */
-    private $entity;
+    private $entityId;
 
     /**
+     * local|admin
+     *
      * @var string
      *
-     * @ORM\Column(type="string", nullable=false)
+     * @ORM\Column(type="string", length=5, nullable=false)
      */
     private $type;
 
     /**
      * @var array
      *
-     * @ORM\Column(type="json_array", nullable=true)
+     * @ORM\Column(type="json_array", nullable=false)
      */
     private $options;
 
     public function __construct(
-        string $entity,
         string $guid,
         string $projectId,
+        string $entityId,
         string $type,
-        DateTime $createdAt,
-        DateTime $updatedAt,
         array $options = []
     ) {
-        $this->entity = $entity;
         $this->guid = $guid;
         $this->projectId = $projectId;
+        $this->entityId = $entityId;
         $this->setType($type);
-        $this->createdAt = $createdAt;
-        $this->updatedAt = $updatedAt;
         $this->options = $options;
     }
 
-    public function getId(): int
+    public function getId(): ?int
     {
         return $this->id;
     }
@@ -100,22 +97,19 @@ class RefOptions
         return $this->projectId;
     }
 
-    /**
-     * @param string $projectId
-     */
     public function setProjectId(string $projectId)
     {
         $this->projectId = $projectId;
     }
 
-    public function getEntity(): string
+    public function getEntityId(): int
     {
-        return $this->entity;
+        return $this->entityId;
     }
 
-    public function setEntity(string $entity): void
+    public function setEntityId(int $entityId): void
     {
-        $this->entity = $entity;
+        $this->entityId = $entityId;
     }
 
     public function getType(): string
@@ -123,24 +117,21 @@ class RefOptions
         return $this->type;
     }
 
-    /**
-     * @param string $type
-     */
-    public function setType(string $type)
+    public function setType(string $type): void
     {
         if (!in_array($type, [self::TYPE_ADMIN, self::TYPE_LOCAL])) {
-            throw new InvalidArgumentException('Type for options not allowed');
+            throw new InvalidArgumentException('Type document for options not allowed');
         }
 
         $this->type = $type;
     }
 
-    public function getOptions(): ?array
+    public function getOptions(): array
     {
         return $this->options;
     }
 
-    public function setOptions(array $options)
+    public function setOptions(array $options = [])
     {
         $this->options = $options;
     }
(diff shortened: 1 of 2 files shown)


src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
     private $id;
 
     /**
-     * @var array
+     * @var string
      *
-     * @ORM\Column(type="string", nullable=false)
+     * @ORM\Column(type="string", length=50, nullable=false)
```
BPScott: Guids are always 36chars long when they're hyphenated

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
-    private $originalId;
+    private $guid;
 
     /**
-     * @var CoreEntity
+     * @var string
      *
-     * @ORM\ManyToOne(targetEntity="CoreEntity")
-     * @ORM\JoinColumn(nullable=true, onDelete="SET NULL")
+     * @ORM\Column(type="string", length=50, nullable=false)
```
BPScott: What is the max length for a project id in iSite? I think it's less than 50

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
      */
-    private $optionsForNetwork;
+    private $entity;
 
     /**
-     * @var array
+     * @var string
      *
-     * @ORM\Column(type="json_array", nullable=true)
+     * @ORM\Column(type="string", nullable=false)
```
BPScott: Add a length so the row size is shorter, it's going to be either 'admin' or 'local' so 5 chars?

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
-        $this->setOriginalId($originalId);
-        $this->setOptionsFor($optionsFor);
+    private $options;
+
+    public function __construct(
+        string $entity,
+        string $guid,
+        string $projectId,
+        string $type,
+        DateTime $createdAt,
```
BPScott: You don't need to pass in createdAt and updatedAt, those properties are handled by the TimestampableEntity trait when you submit to the database.

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
+        $this->guid = $guid;
+        $this->projectId = $projectId;
+        $this->setType($type);
+        $this->createdAt = $createdAt;
+        $this->updatedAt = $updatedAt;
+        $this->options = $options;
     }
 
-    public function getId(): ?int
+    public function getId(): int
```
BPScott: id can be null if you've not yet persisted the entity to the database.

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
+        return $this->projectId;
     }
 
-    public function getOptionsForCoreEntity(): ?CoreEntity
-    {
-        return $this->optionsForCoreEntity;
-    }
-
-    public function getOptionsForNetwork(): ?Network
+    /**
```
BPScott: This docblock doesn't convey any extra info in addition to the typehints that are already in the code so it can be removed.

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +28,120 @@ class RefOptions
...
-    private $optionsForCoreEntity;
+    private $projectId;
 
     /**
-     * @var Network
+     * @var string
      *
-     * @ORM\ManyToOne(targetEntity="Network")
-     * @ORM\JoinColumn(nullable=true, onDelete="SET NULL")
+     * @ORM\Column(type="string", nullable=true)
```
BPScott: What is entity and why is it nullable?

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +27,112 @@ class RefOptions
...
+        }
 
-    public function getProjectSpace(): ?string
-    {
-        return $this->projectSpace;
+        $this->typeDocument = $typeDocument;
     }
 
-    public function setProjectSpace(?string $projectSpace): void
+    public function getOptions(): ?array
```
BPScott: Options seems to be always set, then it doesn't need to be nullable

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -3,14 +3,18 @@
 namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Exception\InvalidArgumentException;
```
BPScott: Should be the base InvalidArgumentException

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +27,112 @@ class RefOptions
...
-     * @var array
+     * local|admin
      *
-     * @ORM\Column(type="json_array", nullable=true)
+     * @var string
+     *
+     * @ORM\Column(type="string", length=5, nullable=false)
      */
-    private $adminOptions;
+    private $typeDocument;
```
BPScott: I reckon call this 'type' we already know it's a document

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -3,14 +3,18 @@
 namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Exception\InvalidArgumentException;
 use Gedmo\Timestampable\Traits\TimestampableEntity;
-use InvalidArgumentException;
 
 /**
  * @ORM\Entity()
+ * @ORM\Table(indexes={@ORM\Index(name="ref_options_idx", columns={"guid", "entity_id"})})
```
BPScott: We need to be able to look things up based on entity id only, so this needs to be two indexs rather than one index with two columns.

If you add `unique=true` to the guid column definition doctrine will add an index for it, so you'd only need to specify the index for the entity_id.

Also check https://github.com/bbc/programmes-pages-service/blob/master/src/Data/ProgrammesDb/Entity/PipsChange.php for how we format this annotation, so it is easy to add new indexs if we ever need to

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +27,112 @@ class RefOptions
...
+     *
+     * @ORM\Column(type="string", length=5, nullable=false)
      */
-    private $adminOptions;
+    private $typeDocument;
 
     /**
      * @var array
      *
      * @ORM\Column(type="json_array", nullable=true)
```
BPScott: Options are always required so should be nullable=false

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -8,9 +8,13 @@
 
 /**
  * @ORM\Entity()
+ * @ORM\Table(indexes={@ORM\Index(name="entity_id_idx", columns={"entity_id"})})
```
BPScott: `ref_options_entity_id_idx` as these index names are global across the DB and it'd be useful to see at-a-glance what table an index refers to

src/Data/ProgrammesDb/Entity/RefOptions.php
```diff
@@ -23,147 +27,112 @@ class RefOptions
...
-     * @ORM\ManyToOne(targetEntity="Network")
-     * @ORM\JoinColumn(nullable=true, onDelete="SET NULL")
-     */
-    private $optionsForNetwork;
-
-    /**
-     * @var array
-     *
-     * @ORM\Column(type="json_array", nullable=true)
+     * @ORM\Column(type="integer", nullable=false)
```
BPScott: entityId would be pid or network ID that comes from isite so it should be a string

PR #117, opened 2017-06-12
Add isRadio() and isTv() methods to programme

BjornTwachtmann:
Any objections? Please say there are no objections.

### DIFF
--- a/src/Domain/Entity/Programme.php
+++ b/src/Domain/Entity/Programme.php
@@ -281,6 +281,24 @@ abstract class Programme
         return $this;
     }
 
+    public function isRadio(): bool
+    {
+        $network = $this->getNetwork();
+        if ($network && $network->isRadio()) {
+            return true;
+        }
+        return false;
+    }
+
+    public function isTv(): bool
+    {
+        $network = $this->getNetwork();
+        if ($network && $network->isTv()) {
+            return true;
+        }
+        return false;
+    }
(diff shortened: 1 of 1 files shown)


src/Domain/Entity/Programme.php
```diff
@@ -281,6 +281,24 @@ public function getTleo(): Programme
         return $this;
     }
 
+    public function isRadio(): bool
+    {
+        $network = $this->getNetwork();
+        if ($network && $network->isRadio()) {
```
votemike: Or even `return $network && $network->isRadio()`?

src/Domain/Entity/Programme.php
```diff
@@ -281,6 +281,24 @@ public function getTleo(): Programme
...
+            return true;
+        }
+        return false;
+    }
+
+    public function isTv(): bool
+    {
+        $network = $this->getNetwork();
+        if ($network && $network->isTv()) {
```
votemike: Or even `return $network && $network-> isTv()`?

votemike:
Do we need tests for this change?

BjornTwachtmann:
Goddammit I said no objections...

votemike:
If you want no objections... Do-It-Right-In-The-First-Place™ 😝

votemike:
Please create a release tag once merged.

PR #118, opened 2017-06-23
PROGRAMMES-5659 added is_sustaining_service into the service entity

Tasiobg:


BjornTwachtmann:
The code-squirrel army approves this PR

PR #119, opened 2017-06-27
New Dependent Block entity, renamed RefOptions, added fileId

felipeparaujo:


BjornTwachtmann:
Sweet. Looks good now. Approved by the army of ninja-code-squirrels.

PR #122, opened 2017-07-05
Isite feature

felipeparaujo:


src/Domain/Entity/Programme.php
```diff
@@ -205,6 +211,22 @@ public function isStreamableAlternatate(): bool
         return $this->isStreamableAlternatate;
     }
 
+    public function getOptions(): Options
```
BPScott: Add an @throws docblock (see getParent below)

src/Mapper/ProgrammesDbToDomain/OptionsMapper.php
```diff
@@ -0,0 +1,70 @@
...
+        $parentEntities[] = $this->getDefaultValues();
+
+        // $parentEntities must start from the bottom of the hierarchy
+
+        // now loop through parents and apply the data
+        // ONLY if the key is allowed to cascade AND
+        // the key doesn't already exist lower down
+        foreach ($parentEntities as $parentOptions) {
+            foreach ($parentOptions as $key => $data) {
+                if (!isset($options[$key]) && (self::OPTIONS_SCHEMA[$key]['cascades'] ?? false)) {
```
BPScott: Is `self::OPTIONS_SCHEMA[$key]['cascades'] ?? false` there  because if an unknown key is found in the options array then it should not cascade?

Could we add a comment to explain that because I'm going to forget that in like 2 weeks

felipeparaujo: will do.

src/Mapper/ProgrammesDbToDomain/OptionsMapper.php
```diff
@@ -0,0 +1,71 @@
...
+        // Append default values to the end of the hierarchy
+        $parentEntities[] = $this->getDefaultValues();
+
+        // $parentEntities must start from the bottom of the hierarchy
+
+        // now loop through parents and apply the data
+        // ONLY if the key is allowed to cascade AND
+        // the key doesn't already exist lower down.
+        // By default, options don't cascade.
+        foreach ($parentEntities as $parentOptions) {
```
BjornTwachtmann: Question about this logic. Should a key get a default value even if it's not set to cascade? I think it probably should, and unless I'm reading this logic wrong it doesn't do that.

felipeparaujo: It doesn't, but it looks like this is what programmes v2 does currently.

felipeparaujo: But you're right, default values should always be assigned.

BjornTwachtmann:
Looks good to me

PR #126, opened 2017-07-21
Adding in a way to get channels

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5698

For https://github.com/bbc/programmes-frontend/pull/99/ Review this first.


votemike:
Added. I guess that's all I could test....

BPScott:
If you're after TV networks (and their default services) that's covered by [NetworksService->findPublishedNetworksByType()](https://github.com/bbc/programmes-pages-service/blob/29dd65a5dcbeb021c42f0be11f5b305c521adf9e/src/Data/ProgrammesDb/EntityRepository/NetworkRepository.php#L21).

You can either filter to only the active ones in PHP (as that list shall be pretty small) or add filter args onto that

PR #130, opened 2017-07-25
Adding core entity service

votemike:


tests/Service/CoreEntityService/AbstractCoreEntityServiceTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $mockCoreEntity->method('getPid')->willReturn(new Pid($entity['pid']));
+        return $mockCoreEntity;
+    }
+
+    protected function service()
+    {
+        return new CoreEntityService($this->mockRepository, $this->mockMapper, $this->mockCache);
+    }
+
+    protected function setUpMapper($mapperName, $entityBuilderMethod)
```
BPScott: This seems to the same as the base class, is it needed?

BPScott:
All our other Services are named after plurals "ContributionsService" "ProgrammesService" etc, we should keep to that consistency.

Also you need to add a method to the ServiceFactory to create this new service.

PR #132, opened 2017-07-31
Clips backend skeleton

tatitati:
Topic: branding page skeleton.


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,31 @@ public function findByIds(array $ids): array
...
+    public function findProgrammesByAncestryAndType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
+    {
+        $this->assertEntityType($entityType, ['Clip', 'Episode', 'Series']);
+
+        $qb = $this->createQueryBuilder('programme')
+                   ->addSelect(['masterBrand', 'image', 'mbImage'])
+                   ->leftJoin('programme.image', 'image')
+                   ->leftJoin('programme.masterBrand', 'masterBrand')
+                   ->leftJoin('masterBrand.image', 'mbImage')
+                   ->where("programme INSTANCE OF ProgrammesPagesService:$entityType")
```
felipeparaujo: This should be using a parameter

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,33 @@
...
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    public function findChildrensByType(
+        array $findByAncestryIds,
+        string $type,
+        ?int $limit = self::DEFAULT_LIMIT,
+        int $page = self::DEFAULT_PAGE
+    ) {
+        return $this->repository->findProgrammesByAncestryAndType(
```
felipeparaujo: This should be using the cache stuff the other services use as well. Also, any specific reason for this not to be included in the programmes service?

src/Service/ServiceFactory.php
```diff
@@ -223,4 +223,17 @@ public function getVersionsService(): VersionsService
 
         return $this->instances['VersionsService'];
     }
+
+    public function getClipsService(): ProgrammesAggregationService
+    {
+        if (!isset($this->instances['ClipsService'])) {
```
felipeparaujo: If this is an instance of ProgrammesAggregationService, why is this not called getProgrammesAggregationService?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,31 @@ public function findByIds(array $ids): array
...
+                   ->andWhere('programme.ancestry LIKE :ancestry')
+                   ->andWhere('programme.streamable = 1')
+                   ->setFirstResult($offset)
+                   ->setMaxResults($limit)
+                   ->setParameter('ancestry', $this->ancestryIdsToString($ancestryDbIds) . '%');
+
+        if (in_array($entityType, ['Episode', 'Clip'])) {
+               // for programmeItems(clips, episode) return results ordered by streamableFrom
+               $qb->leftJoin("BBC\\ProgrammesPagesService\\Data\\ProgrammesDb\\Entity\\$entityType", 'programmeItem', 'WITH', 'programme = programmeItem');
```
felipeparaujo: This should also be using a parameter

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,31 @@ public function findByIds(array $ids): array
             ->getQuery()->getResult(Query::HYDRATE_ARRAY);
     }
 
+    public function findProgrammesByAncestryAndType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
+    {
+        $this->assertEntityType($entityType, ['Clip', 'Episode', 'Series']);
```
votemike: Are 'Clip', 'Episode' and 'Series' not constants anywhere?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
votemike: Is it safe to put $entityType in like this?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,31 @@ public function findByIds(array $ids): array
             ->getQuery()->getResult(Query::HYDRATE_ARRAY);
     }
 
+    public function findProgrammesByAncestryAndType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
```
votemike: What's `$ancestryDbIds` an array of? ints?

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    public function __construct(
+        CoreEntityRepository $repository,
+        CoreEntityMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    public function findChildrensByType(
```
votemike: `findChildrenByType()`

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Service;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ClipRepository;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\CoreEntityRepository;
+use BBC\ProgrammesPagesService\Mapper\ProgrammesDbToDomain\CoreEntityMapper;
+use BBC\ProgrammesPagesService\Cache\CacheInterface;
+
+class ProgrammesAggregationService extends AbstractService
```
votemike: What is this class for?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: It should be safe, as we're checking the parameter beforehand against a known whitelist. It's not ideal, but it's what everything else in the class also does.

src/Service/ProgrammesAggregationService.php
votemike: Would it not be better to put these methods in the CoreEntitiesService?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: I don't think you can use a parameter there. Everything else in this class also does this.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: Why in the name of Satan's right testicle are you leftJoining core_entity to itself? That makes no sense to me.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
tatitati: no yet

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
tatitati: trying to access to the property "streamableFrom" by using programmes make Doctrine complains (streamableFrom property only exist for the subclass programmeItems). So it doesn't allow me to filter by that property (specific of a subclass). I found this solution on stack overflow: https://stackoverflow.com/questions/8496621/doctrine-dql-class-table-inheritance-and-access-to-subclass-fields

Anyone with a better idea?.

BjornTwachtmann: Ugh... Doctrine. A self left-join here will have *serious* performance implications (assuming doctrine is actually representing it that way in the SQL?). It's not acceptable. Will come over and take a look.

src/Service/ProgrammesAggregationService.php
tatitati: Discussing with some people about it we thought that if we keep adding all the stuff in CoreEntitiesService this will grow up too much. But happy to move it back to CoreEntities if necessary.

src/Service/ProgrammesAggregationService.php
tatitati: thanks!, very appreciated

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
tatitati: yeah I know it has an impact. Thanks for the help.

src/Service/ProgrammesAggregationService.php
votemike: @BPScott Any thoughts?

src/Service/ProgrammesAggregationService.php
BPScott: `ProgrammeAggregationsService` not, `ProgrammesAggregationService`. The whole thing is a plural.

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,64 @@
...
+    public function __construct(
+        CoreEntityRepository $repository,
+        CoreEntityMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    public function findChildrenByType(
+        array $findByAncestryIds,
```
felipeparaujo: This should be the entity itself, and you should get the ancestry ids in this function

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
```diff
@@ -0,0 +1,39 @@
...
+    public function testFindEpisodeGuideChildrenGetCorrectEpisodes()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+
+        $episodesUnderProgramme = $this->getEntityManager()
+                                    ->getRepository('ProgrammesPagesService:CoreEntity')
+                                    ->findProgrammesByAncestryAndType([1, 14], 'Episode', 100, 0);
+
+
+        $this->assertCount(3, $episodesUnderProgramme);
```
felipeparaujo: Could you be a bit more specific in these tests? just check the pids to make sure you're getting the clips you're expecting in the order you're expecting them

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+class FindProgrammesByAncestryAndTypesTest extends AbstractDatabaseTest
+{
+    public function testFindEpisodeGuideChildrenGetCorrectClips()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+
+        $clipsUnderProgramme = $this->getEntityManager()
+            ->getRepository('ProgrammesPagesService:CoreEntity')
+            ->findProgrammesByAncestryAndType([1, 15], 'Clip', 100, 0);
```
votemike: What's 1 and 15?

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
```diff
@@ -0,0 +1,39 @@
...
+use Tests\BBC\ProgrammesPagesService\AbstractDatabaseTest;
+
+/**
+ * @covers BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\CoreEntityRepository::<public>
+ * @group repository
+ * @group programmesAggregation
+ */
+class FindProgrammesByAncestryAndTypesTest extends AbstractDatabaseTest
+{
+    public function testFindEpisodeGuideChildrenGetCorrectClips()
```
votemike: At first glance, I don't quite understand the test name. What is this test doing?

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
votemike: And make sure they are all instances of `Episode?`

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
tatitati: ancestryIds

felipeparaujo: yeah this could lead to potentially flaky tests. Take a look at [this](https://github.com/bbc/programmes-pages-service/blob/008ab25f031c3c389dbe6e7fa485952cacef1459/tests/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository/CountByCategoryAncestryAndEndAtDateRangeTest.php#L27). Use that getAncestry function to get the ancestry instead of using the ids straight away

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
```diff
@@ -0,0 +1,55 @@
...
+ * @group repository
+ * @group programmesAggregation
+ */
+class FindProgrammesByAncestryAndTypesTest extends AbstractDatabaseTest
+{
+    public function testProgrammesChildrenGetCorrectClips()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+
+        $dbAncestryId = [1, 15];
```
felipeparaujo: yeah this could lead to potentially flaky tests. Take a look at this. Use that getAncestry function to get the ancestry instead of using the ids straight away

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,38 @@ public function findByIds(array $ids): array
...
+            ->setMaxResults($limit)
+            ->setParameter('ancestry', $this->ancestryIdsToString($ancestryDbIds) . '%');
+
+        if (in_array($entityType, ['Episode', 'Clip'])) {
+            // Doctrine will complain loudly if the entity does not have the
+            // field to order by on it.
+            $qb->orderBy('programme.streamableFrom', 'DESC');
+        }
+
+        return $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
```
BPScott: You'll need to resolve the ancestry here otherwise we won't be able to build the programme objects later.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,38 @@ public function findByIds(array $ids): array
...
     }
 
+    /**
+     * @param int[]    $ancestryDbIds
+     * @param string   $entityType
+     * @param int|null $limit
+     * @param int      $offset
+     * @return array
+     */
+    public function findProgrammesByAncestryAndType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
```
BPScott: FindDescendantsByType might be a better name

I'd be tempted to open this up to all types now, to save you having to edit this when you add in group support later, if you do that then rename 'programme' to be 'entity' because it could be a programme or group .

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,68 @@
...
+        CoreEntityMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    /**
+     * @return Programme[]
+     */
+    public function findProgrammesChildrenByType(
```
BPScott: How about `findDescendantClips` - then other filters can have separate names `findDescendantEpisodes` etc. That shall help discoverability in your IDE as your function names make it clear what things do.


Note that this is in contrast o the Repository functions - the Repositories are private - they don't need to be accessed outside of this library. This means that we should focus on reducing the number of function by reusing / finding common ground where appropriate, rather than developer experience.

Services - public - DX means discoverability by IDE hints on function names and not having to guess parameter values.
Repositories - private - DX means keeping a small but flexible footprint when making things that the Services use.

src/Service/ServiceFactory.php
```diff
@@ -223,4 +223,17 @@ public function getVersionsService(): VersionsService
 
         return $this->instances['VersionsService'];
     }
+
+    public function getProgrammesAggregationService(): ProgrammesAggregationService
```
BPScott: Keep the functions in this (any all other) Factories alphabetical so it is easy to find them

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindProgrammesByAncestryAndTypesTest.php
```diff
@@ -0,0 +1,56 @@
...
+    public function testProgrammesChildrenGetCorrectClips()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+        $dbAncestryId = $this->getAncestryFromPersistentIdentifier('b010t19z', 'Brand');
+
+        $clipsUnderProgramme = $this->getEntityManager()
+            ->getRepository('ProgrammesPagesService:CoreEntity')
+            ->findProgrammesByAncestryAndType($dbAncestryId, 'Clip', 100, 0);
+
+        $this->assertCount(5, $clipsUnderProgramme);
```
BPScott: Check out using the `$this->assertEquals($expectedPids, array_column($entities, 'pid'));` style from the other tests, e.g. FindChildrenSeriesByParentTest.

That will save you testing each row of the results.

tests/Service/ProgrammesAggregationService/FindProgrammesChildrenByTypeTest.php
```diff
@@ -0,0 +1,39 @@
...
+ * @covers BBC\ProgrammesPagesService\Service\ProgrammesAggregationService::findProgrammesChildrenByType
+ * @group service
+ * @group programmesAggregation
+ */
+class FindProgrammesChildrenByTypeTest extends AbstractProgrammesAggregationTest
+{
+    /**
+     * @expectedException InvalidArgumentException
+     */
+    public function testCannotGetBrandTypesChildrenForAnyAncestry()
```
BPScott: I tend to put the error tests towards the bottom - making sure stuff works is the important thing to look for as you're scrolling through

tests/Service/ProgrammesAggregationService/FindProgrammesChildrenByTypeTest.php
```diff
@@ -0,0 +1,39 @@
...
+        );
+    }
+
+    public function testFindChildrenByTypeCallsProperMethods()
+    {
+        $mockProgramme = $this->createMock(Programme::class);
+        $mockProgramme->method('getDbAncestryIds')->willReturn([11, 12]);
+
+        $this->mockRepository->expects($this->once())
+                             ->method('findProgrammesByAncestryAndType')
```
BPScott: code style is don't do the deep indent thing. Only indent 4 spaces.

tests/Service/ProgrammesAggregationService/FindProgrammesChildrenByTypeTest.php
```diff
@@ -0,0 +1,39 @@
...
+                             ->willReturn([['type' => 'clip'], ['type' => 'clip']]);
+
+        $mappedResults = $this->service()->findProgrammesChildrenByType($mockProgramme, 'Clip');
+        $this->assertContainsOnlyInstancesOf('BBC\ProgrammesPagesService\Domain\Entity\Clip', $mappedResults);
```
BPScott: Use `$this->programmesFromDbData($dbData)` like in `FindChildrenSeriesByParentTest`

src/Service/ProgrammesAggregationService.php
BPScott: I'm a little wary of piling everything into CoreEntitesService / ProgrammesService as then they end up getting HUUUUUUUUGE and unwieldy. I'm pondering the idea of having multiple Services that can all return the same type of thing. More obvious examples would be `ProgrammesAtoZService` and `ProgrammesCategoriesService` that contain service calls related to getting a-z and per-category filtered programmes.

I know I mentioned `ProgrammesAggregationService` first but perhaps something like `ProgrammesByAncestorService` might be a bit clearer with the intent - to contains service calls get a list of programmes by some common ancestor. e.g. "get all clips beneath programme X" or "get all episodes that are direct children of programme X"

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $expectedClipsPids = ['p00h64pq', 'p00hv9yz', 'p008k0l5', 'p008k0jy', 'p008nhl4'];
+        $this->assertEquals($expectedClipsPids, array_column($clipsUnderProgramme, 'pid'));
+        // all clips are streamable
+        $this->assertEquals(
+            [true],
+            array_unique(array_column($clipsUnderProgramme, 'streamable'))
+        );
+        // all clips are not embargoed
+        $this->assertEquals(
```
BPScott: `$this->assertNotContains(true, array_column($clipsUnderProgramme, 'isEmbargoed'));` might be a bit shorter and more readable

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $dbAncestryId = $this->getAncestryFromPersistentIdentifier('b010t19z', 'Brand');
+
+        $clipsUnderProgramme = $this->getEntityManager()
+            ->getRepository('ProgrammesPagesService:CoreEntity')
+            ->FindDescendantsByType($dbAncestryId, 'Clip', 100, 0);
+
+        $expectedClipsPids = ['p00h64pq', 'p00hv9yz', 'p008k0l5', 'p008k0jy', 'p008nhl4'];
+        $this->assertEquals($expectedClipsPids, array_column($clipsUnderProgramme, 'pid'));
+        // all clips are streamable
+        $this->assertEquals(
```
BPScott: `$this->assertNotContains(false, array_column($clipsUnderProgramme, 'streamable'));` might be a bit shorter and more readable

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->loadFixtures(['MongrelsFixture']);
+        $dbAncestryId = $this->getAncestryFromPersistentIdentifier('b010t19z', 'Brand');
+
+        $episodesUnderProgramme = $this->getEntityManager()
+                                    ->getRepository('ProgrammesPagesService:CoreEntity')
+                                    ->FindDescendantsByType($dbAncestryId, 'Episode', 100, 0);
+
+        $expectedClipsPids = ['b00tf1zy', 'b00swgkn', 'b00syxx6', 'b00t0ycf', 'b0175lqm', 'b0176rgj', 'b0177ffr'];
+        $this->assertEquals($expectedClipsPids, array_column($episodesUnderProgramme, 'pid'));
+        $this->assertEquals(
```
BPScott: `$this->assertNotContains()` as above?

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+            [true],
+            array_unique(array_column($episodesUnderProgramme, 'streamable'))
+        );
+        $this->assertEquals(
```
BPScott: `$this->assertNotContains()` as above?

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->assertCount(2, $this->getDbQueries());
+    }
+
+    public function testProgrammesChildrenGetCorrectSeries()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+        $dbAncestryId = $this->getAncestryFromPersistentIdentifier('b010t19z', 'Brand');
+
+        $seriesUnderProgramme = $this->getEntityManager()
+                                    ->getRepository('ProgrammesPagesService:CoreEntity')
```
BPScott: Indentation. We always do 4 spaces only.

tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindDescendantsByTypeTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $this->assertCount(2, $this->getDbQueries());
+    }
+
+    public function testProgrammesChildrenGetCorrectEpisodes()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+        $dbAncestryId = $this->getAncestryFromPersistentIdentifier('b010t19z', 'Brand');
+
+        $episodesUnderProgramme = $this->getEntityManager()
+                                    ->getRepository('ProgrammesPagesService:CoreEntity')
```
BPScott: Indentation

tests/Service/ProgrammesAggregationService/FindDescendantClipsTest.php
```diff
@@ -0,0 +1,25 @@
...
+class FindDescendantClipsTest extends AbstractProgrammesAggregationTest
+{
+    public function testFindDescendantClips()
+    {
+        $mockProgramme = $this->createMock(Programme::class);
+        $mockProgramme->method('getDbAncestryIds')->willReturn([11, 12]);
+
+        $this->mockRepository->expects($this->once())
+             ->method('findDescendantsByType')
+             ->willReturn([['type' => 'clip'], ['type' => 'clip']]);
```
BPScott: You're missing an assert on the arguments that get passed to `findDescendantsByType`.

Also please use the `$this->programmesFromDbData($dbData)` pattern for generating and testing the results like in every other Service test.

tests/Service/ProgrammesAggregationService/FindDescendantEpisodesTest.php
```diff
@@ -0,0 +1,25 @@
...
+class FindDescendantEpisodesTest extends AbstractProgrammesAggregationTest
+{
+    public function testFindDescendantEpisodes()
+    {
+        $mockProgramme = $this->createMock(Programme::class);
+        $mockProgramme->method('getDbAncestryIds')->willReturn([11, 12]);
+
+        $this->mockRepository->expects($this->once())
+                             ->method('findDescendantsByType')
+                             ->willReturn([['type' => 'episode'], ['type' => 'episode']]);
```
BPScott: You're missing an assert on the arguments that get passed to `findDescendantsByType`.

Also please use the `$this->programmesFromDbData($dbData)` pattern for generating and testing the results like in every other Service test.

tests/Service/ProgrammesAggregationService/FindDescendantSeriesTest.php
```diff
@@ -0,0 +1,25 @@
...
+class FindDescendantSeriesTest extends AbstractProgrammesAggregationTest
+{
+    public function testFindDescendantSeries()
+    {
+        $mockProgramme = $this->createMock(Programme::class);
+        $mockProgramme->method('getDbAncestryIds')->willReturn([11, 12]);
+
+        $this->mockRepository->expects($this->once())
+                             ->method('findDescendantsByType')
+                             ->willReturn([['type' => 'series'], ['type' => 'series']]);
```
BPScott: You're missing an assert on the arguments that get passed to `findDescendantsByType`.

Also please use the `$this->programmesFromDbData($dbData)` pattern for generating and testing the results like in every other Service test.

src/Service/ProgrammesAggregationService.php
```diff
@@ -26,23 +31,56 @@ public function __construct(
     }
 
     /**
-     * @return Programme[]
+     * @return Series[]
      */
-    public function findProgrammesChildrenByType(
+    public function findDescendantSeries(
```
BPScott: Do we actually do this for lists of Series in frontend? http://www.bbc.co.uk/programmes/b006q2x0/series only shows immediate child series, not from all descendants.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,47 @@ public function findByIds(array $ids): array
...
     }
 
+    /**
+     * @param int[]    $ancestryDbIds
+     * @param string   $entityType
+     * @param int|null $limit
+     * @param int      $offset
+     * @return array
+     */
+    public function findDescendantsByType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
```
votemike: Is this supposed to be public?

votemike: Should this have a space before the colon? E.G. `): array` instead of `) : array`. If so, should that be added to our codesniffer?

tatitati: yeah, I really want to update our codesniffer rules to hightlights this kind of things. I will remember this case as well, I was trying to collect all the cases we want to cover and I didn't see this case.

tatitati: yes, is used in ProgrammesAggregationService.php in multiple cases:
https://github.com/bbc/programmes-pages-service/blob/21b5352489199a35032b314d8f0822ed71057820/src/Service/ProgrammesAggregationService.php#L41

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -213,6 +213,8 @@ public function findByIds(array $ids): array
     public function findDescendantsByType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
     {
         $this->assertEntityType($entityType, ['Clip', 'Episode', 'Series', 'Franchise', 'Gallery', 'Collection', 'Season']);
+        $streamableEntities = ['Episode', 'Clip', 'Series'];
```
BPScott: Brands too.

tatitati: Yeah, I asked to @BjornTwachtmann about it. Also I saw this in pages-services and I tested it:
https://github.com/bbc/programmes-pages-service/blob/master/src/Data/ProgrammesDb/Entity/Programme.php#L50

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Service;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ClipRepository;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\CoreEntityRepository;
+use BBC\ProgrammesPagesService\Domain\Entity\CoreEntity;
+use BBC\ProgrammesPagesService\Domain\Entity\Programme;
+use BBC\ProgrammesPagesService\Mapper\ProgrammesDbToDomain\CoreEntityMapper;
+use BBC\ProgrammesPagesService\Cache\CacheInterface;
```
votemike: In PHPStorm could you run "Optimize Imports" on this file?

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,117 @@
...
+use InvalidArgumentException;
+
+class ProgrammesAggregationService extends AbstractService
+{
+    const AGGREGATION_VALID_TYPES = [
+        'Series',
+        'Episode',
+        'Clip',
+        'Franchise',
```
votemike: You don't have `findDescendantFranchise()`, `findDescendantGallery()`, `findDescendantCollection()` or `findDescendantSeason()` methods. Is that correct?

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,117 @@
...
+     * @return CoreEntity[]
+     */
+    private function findDescendantsByType(
+        Programme $programme,
+        string $type,
+        ?int $limit,
+        int $page,
+        $ttl = CacheInterface::NORMAL
+    ) : array {
+        $this->assertEntityType($type, self::AGGREGATION_VALID_TYPES);
```
votemike: Do you need to pass `self::AGGREGATION_VALID_TYPES` in? Can't you just reference it in `assertEntityType()`?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
votemike: Are you updating the sniffer for PPS?

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,98 @@
...
+     * @return ProgrammeItem[]
+     */
+    private function findStreamableDescendantsByType(
+        Programme $programme,
+        string $type,
+        ?int $limit,
+        int $page,
+        $ttl = CacheInterface::NORMAL
+    ): array {
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $type, $limit, $page, $ttl);
```
votemike: This key needs the $programme ID or something doesn't it?

src/Service/ProgrammesAggregationService.php
```diff
@@ -0,0 +1,98 @@
...
+     * @return ProgrammeContainer[]
+     */
+    private function findNoStreamableDescendantsByType(
+        Programme $programme,
+        string $type,
+        ?int $limit,
+        int $page,
+        $ttl = CacheInterface::NORMAL
+    ): array {
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $type, $limit, $page, $ttl);
```
votemike: This key needs the $programme ID or something doesn't it?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,54 @@ public function findByIds(array $ids): array
             ->getQuery()->getResult(Query::HYDRATE_ARRAY);
     }
 
+    public function findDescendantsByType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
+    {
+        $qb = $this->getEntityManager()->createQueryBuilder()
```
votemike: Do we want to make sure that the $entityType isn't a Clip or Episode? @BjornTwachtmann

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,54 @@ public function findByIds(array $ids): array
...
+                   ->setFirstResult($offset)
+                   ->setMaxResults($limit)
+                   ->setParameter('ancestry', $this->ancestryIdsToString($ancestryDbIds) . '%');
+
+        $result = $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
+        return $this->resolveParents($result);
+    }
+
+    /**
+     * @param int[]    $ancestryDbIds
```
votemike: Add this docblock to findDescendantsByType() too?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -203,6 +203,54 @@ public function findByIds(array $ids): array
             ->getQuery()->getResult(Query::HYDRATE_ARRAY);
     }
 
+    public function findDescendantsByType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
+    {
+        $qb = $this->getEntityManager()->createQueryBuilder()
+                   ->addSelect(['entity', 'masterBrand', 'image', 'mbImage', 'network'])
```
votemike: This is all indented too far.

BjornTwachtmann:
Fix the broken test and then merge this.

PR #134, opened 2017-08-07
Adding useful methods for CoreEntity

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5786

### DIFF
--- a/src/Domain/Entity/Brand.php
+++ b/src/Domain/Entity/Brand.php
@@ -4,5 +4,5 @@ namespace BBC\ProgrammesPagesService\Domain\Entity;
 
 class Brand extends ProgrammeContainer
 {
-    const TYPE = 'brand';
+    protected const TYPE = 'brand';
 }
--- a/src/Domain/Entity/Brand.php
+++ b/src/Domain/Entity/Brand.php
@@ -4,4 +4,5 @@ namespace BBC\ProgrammesPagesService\Domain\Entity;
 
 class Brand extends ProgrammeContainer
 {
+    protected $type = 'brand';
 }
--- a/src/Domain/Entity/Brand.php
+++ b/src/Domain/Entity/Brand.php
@@ -4,5 +4,5 @@ namespace BBC\ProgrammesPagesService\Domain\Entity;
 
 class Brand extends ProgrammeContainer
 {
-    protected $type = 'brand';
+    const TYPE = 'brand';
 }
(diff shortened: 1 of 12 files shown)


src/Domain/Entity/Group.php
```diff
@@ -4,5 +4,13 @@
 
 abstract class Group extends CoreEntity
 {
+    public function isTlec(): bool
+    {
+        return true;
```
BPScott: Groups are not ProgrammeContainers, so they're not Tlecs, this can be removed

src/Domain/Entity/Group.php
```diff
@@ -4,5 +4,13 @@
...
+    }
 
+    public function isTleo(): bool
+    {
+        return true;
```
BPScott: We've now got a setup where Groups may have parents (e.g. a Gallery may belong to an Episode) so they should follow the same logic as everything else - check the parent, so this can be removed

src/Domain/Entity/CoreEntity.php
```diff
@@ -18,6 +18,9 @@
     /** @var Pid */
     protected $pid;
 
+    /** @var string */
+    protected $type = '';
```
BPScott: What do you reckon to these being protected class constants? Then we wouldn't end up creating new strings for every entity we create (micro-memory optimisations)

src/Domain/Entity/Group.php
votemike: But https://github.com/bbc/programmes/blob/636e751443c3d6504282fd1d5c03eb202a235c61/webapp/php/sharedlib/BBC/Programmes/Group.php#L332

src/Domain/Entity/CoreEntity.php
votemike: Could do.... Is it worth it?

src/Domain/Entity/Group.php
BjornTwachtmann: Do we ever consider groups TLEOs? It's not really a PIPs concept so we can define it as we'd like, but I'd say groups can't be TLEOs or TLECs the way we normally define those

votemike: Leave this method in? Take it out?

BjornTwachtmann: Probably should just return false. In fact, a group cannot be the parent of a programme so it's utterly academic anyway.

src/Domain/Entity/Brand.php
```diff
@@ -4,5 +4,5 @@
 
 class Brand extends ProgrammeContainer
 {
-    protected $type = 'brand';
+    const TYPE = 'brand';
```
BPScott: From php7.1 we can specify the visibility so we don't expose these constants publicly `protected const TYPE = 'brand';`

PR #136, opened 2017-08-09
Model promotions

tatitati:
First PR for ticket:
'Build Domain Models and DB->Domain Mapper for Promotions'

This PR only includes the domain entity for promotions.

### DIFF
--- a/src/Domain/Entity/Promotion.php
+++ b/src/Domain/Entity/Promotion.php
@@ -7,9 +7,6 @@ use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
 
 class Promotion
 {
-    /** @var bool */
-    private $cascadesToDescendants;
-
     /** @var Pid */
     private $pid;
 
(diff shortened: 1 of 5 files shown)


src/Domain/Entity/Promotion.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->cascadesToDescendants = $cascadesToDescendants;
+        $this->pid = $pid;
+        $this->promotedEntity = $promotedEntity;
+        $this->synopses = $synopses;
+        $this->title = $title;
+        $this->url = $url;
+        $this->weighting = $weighting;
+    }
+
+    public function isCascadesToDescendants(): bool
```
votemike: The method name doesn't quite make sense. But can't think of a better one at the moment.

tatitati: well, I thought that we can use something like "isPropagated", or "isSuperpromotion (we call them in this way when they are propagated), but "isCascade" is also  familiar to the field we have in database, so some people is familiar and means pretty much the same.....

someone with any thought?

BPScott: I don't think we'd ever need to call this from code, we only use it for filtering in DB queries, so I don't think there's any value having it here at all.

BjornTwachtmann:
I don't think it's worth including the date. If we're showing a promo for the extra 2 minutes it's cached in redis... So what. A lot of DateTime logic is expensive and annoying to maintain.

BPScott:
Cool, ignore that thought then :)

PR #137, opened 2017-08-09
Add option to inherit MasterBrand from parent CoreEntity

BPScott:
It is possible that a CoreEntity that is not relatated to a MasterBrand
does have a parent that is related to a MasterBrand, for example p043m65r

In the new frontend we want these Entities that do not have an explict
MasterBrand set to try and find one from their ancestry, in the same way
that we try and inherit images.

Clifton however does not do this inheritance.

In order to not make changes to Clifton's output, add an option
"core_entity_inherit_master_brand" to the MapperFactory that allows us
to choose between these two behaviours. When consuming PagesService you
probably want to set this option to true. We have not set this to be the
default as that would constitiute a breaking change, that we don't want
to make right now.

### DIFF
--- a/src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
+++ b/src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
@@ -493,11 +493,42 @@ class CoreEntityMapper extends AbstractMapper
             return new UnfetchedMasterBrand();
         }
 
-        if (is_null($dbEntity[$key])) {
+        // CoreEntities may not have MasterBrands.
+        // Most systems will want to look up the parent hierarchy to see if
+        // there is a MasterBrand we can attach to the current item. However
+        // some legacy APIs we still need to maintain (e.g. Clifton) do not
+        // expose the inherited MasterBrand.
+        if (!$this->mapperFactory->getOption('core_entity_inherit_master_brand')) {
+            return $this->getMasterBrandModelWithoutInheritance($dbEntity[$key]);
+        }
+
+        $masterBrandMapper = $this->mapperFactory->getMasterBrandMapper();
+
+        // MasterBrand inheritance. If the current entity does not have a
+        // masterbrand attached to it, look to see if its parent has a
+        // masterbrand, and use that. Keep going up the ancestry chain till a
+        // masterbrand is found
+        $currentItem = $dbEntity;
+        while ($currentItem) {
+            // If the current Entity has a masterbrand then use that!
+            if (isset($currentItem[$key])) {
+                return $masterBrandMapper->getDomainModel($currentItem[$key]);
+            }
+
+            // Otherwise set the current Entity to the parent
+            $currentItem = $currentItem['parent'] ?? null;
+        }
+
+        return null;
+    }
+
+    private function getMasterBrandModelWithoutInheritance(?array $dbMasterBrand): ?MasterBrand
+    {
+        if (is_null($dbMasterBrand)) {
             return null;
         }
 
-        return $this->mapperFactory->getMasterBrandMapper()->getDomainModel($dbEntity[$key]);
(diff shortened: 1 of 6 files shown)


BPScott:
This is the first step to fixing PROGRAMMES-5787

src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
```diff
@@ -493,11 +493,41 @@ private function getMasterBrandModel(array $dbEntity, string $key = 'masterBrand
             return new UnfetchedMasterBrand();
         }
 
-        if (is_null($dbEntity[$key])) {
+        // CoreEntities may not have masterbrands.
+        // In the new front-end we want to look up the parent hierarchy to
```
votemike: "new front-end" or "v3"?

src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
```diff
@@ -493,11 +493,41 @@ private function getMasterBrandModel(array $dbEntity, string $key = 'masterBrand
...
+        // see if there is a masterbrand we can attach do the current item.
```
votemike: "can attach do" => "can attach to"

src/Mapper/ProgrammesDbToDomain/MapperFactory.php
```diff
@@ -4,8 +4,29 @@
 
 class MapperFactory
 {
+    /** @var array */
     protected $instances = [];
 
+    /** @var array */
+    private $options = [
+        // In Clifton the MasterBrand is the one associated with the current
```
votemike: PPS is just a system agnostic package added via Composer. Should such a package mention specific use cases like "Clifton" or "Frontend"?

tests/Mapper/ProgrammesDbToDomain/CoreEntityMapperMasterBrandMappingTest.php
```diff
@@ -35,10 +35,72 @@ public function testGetDomainModelSeriesWithSetMasterBrand()
             $expectedMasterBrandDomainEntity
         );
 
-        $this->assertEquals($expectedEntity, $this->getMapper()->getDomainModelForProgramme($dbEntityArray));
+        $this->assertEquals($expectedEntity, $this->getMapper($options)->getDomainModelForProgramme($dbEntityArray));
+    }
+
+    public function masterBrandOnCurrentEntityDataProvider()
```
votemike: return type

tests/Mapper/ProgrammesDbToDomain/CoreEntityMapperMasterBrandMappingTest.php
```diff
@@ -35,10 +35,72 @@ public function testGetDomainModelSeriesWithSetMasterBrand()
...
+    {
+        return [
+            [[], $this->createMock(MasterBrand::class)], // default value is true
```
votemike: Could you add array keys describing the tests so that when they fail we know which it is.

tests/Mapper/ProgrammesDbToDomain/CoreEntityMapperOptionsMappingTest.php
```diff
@@ -27,7 +27,7 @@ public function testGetDomainModelHierarchy()
 
         $expectedMasterBrandDomainEntity = $this->createMock(MasterBrand::class);
 
-        $this->mockMasterBrandMapper->expects($this->once())
+        $this->mockMasterBrandMapper->expects($this->atLeastOnce())
```
votemike: Why has this needed to change?

BPScott: This test contains a programme with a parent. Now that "try look up your masterbrand from your parent" is the default, then the masterBrand mapper is called twice - one for the itself, and one for the parent element.

PR #139, opened 2017-08-10
Use a cache to make a time/memory tradeoff and reduce new Cronos crea…

BjornTwachtmann:
This reduces response time for FE by a couple of ms. Not much, but easy to do.

### DIFF
--- a/src/Data/ProgrammesDb/Type/UtcDateTimeType.php
+++ b/src/Data/ProgrammesDb/Type/UtcDateTimeType.php
@@ -13,6 +13,8 @@ use InvalidArgumentException;
 
 class UtcDateTimeType extends DateTimeType
 {
+    private static $cache = [];
+
     public function convertToDatabaseValue($value, AbstractPlatform $platform): ?string
     {
         if (is_null($value)) {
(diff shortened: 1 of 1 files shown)


src/Data/ProgrammesDb/Type/UtcDateTimeType.php
```diff
@@ -13,6 +13,8 @@
 
 class UtcDateTimeType extends DateTimeType
 {
+    private static $cache = [];
```
votemike: A docblock about what this is... I dunno if `Chronos[][]` works or if it would be `array[]`

src/Data/ProgrammesDb/Type/UtcDateTimeType.php
```diff
@@ -38,17 +40,17 @@ public function convertToPHPValue($value, AbstractPlatform $platform): ?Chronos
...
-        $converted = Chronos::createFromFormat($platform->getDateTimeFormatString(), $value, 'UTC');
-
-        if (!$converted) {
-            throw ConversionException::conversionFailedFormat(
-                $value,
-                $this->getName(),
-                $platform->getDateTimeFormatString()
-            );
+        $dateFormatString = $platform->getDateTimeFormatString();
+        if (!isset(self::$cache[$dateFormatString][$value])) {
```
votemike: Do we want to ensure $value is a string before this point?

BjornTwachtmann: We've already checked for null. It's going to be null or a string. This blows up with an exception if its not anyway.

PR #141, opened 2017-08-10
Phpcs coding style symfony3

tatitati:
Migrate phpcs standard to symfony 3

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/Traits/ParentTreeWalkerTrait.php
+++ b/src/Data/ProgrammesDb/EntityRepository/Traits/ParentTreeWalkerTrait.php
@@ -116,7 +116,7 @@ trait ParentTreeWalkerTrait
      *
      * @return array|null
      */
-    private function combineAncestry($entity, array $potentialAncestors = [], $keyPath = ['ancestry'])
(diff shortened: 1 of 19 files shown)


src/Data/ProgrammesDb/EntityRepository/Traits/ParentTreeWalkerTrait.php
```diff
@@ -114,12 +114,9 @@ protected function abstractResolveNestedAncestry(array $entities, string $key, c
     /**
      * Using the potential ancestors as a source, recursively set the parents into place
      *
-     * @param $entity
```
votemike: Is there a reason you have removed the params but left the return value?

votemike: Instead of removing these, could you typehint in them?

tatitati: well, I removed them because phpcs complains when you add a @param item without the type. Yeah, I will. :)

BjornTwachtmann:
I am extremely happy that the number of spaces between things is now consistent. I cannot tell everybody how much this issue has been troubling me...

Looks good mate, just merge it.

PR #142, opened 2017-08-11
Mapper improvements

BPScott:
* Adding tests for building Groups in CoreEntityMapper
* Refactoring how we get get models for CoreEntities to avoid having dead code and duplicate isset checks


tatitati:
nice

votemike:
Looks OK to me.

PR #143, opened 2017-08-14
Use the same CacheKey for all CoreEntities

BPScott:
There is no value in having different keys for Programmes vs Groups as
the only value that differs is that Programmes may be joined to
categories.

This removes the need for a few isset() checks, and simplifies the
public API by removing the getCacheKeyForProgramme and
getCacheKeyForGroup methods.


votemike:
It doesn't say this was merged. How come?

BPScott:
A bit of rebase wrangling confusing github. Pushing the rebased commit to master, before this branch. https://github.com/bbc/programmes-pages-service/commits/master shows it in master

PR #144, opened 2017-08-14
promotion domain tests

tatitati:
I forgot to add tests for promotion entity domain.


src/Domain/Entity/Promotion.php
```diff
@@ -51,6 +51,11 @@ public function getPromotedEntity(): PromotableInterface
         return $this->promotedEntity;
     }
 
+    public function getSynopses()
```
BPScott: Typehint?

PR #145, opened 2017-08-14
promotions-mapper

tatitati:
once that promotions-repository is ready then the mapper is necessary (THIS PR)

### DIFF
--- a/src/Mapper/ProgrammesDbToDomain/MapperFactory.php
+++ b/src/Mapper/ProgrammesDbToDomain/MapperFactory.php
@@ -114,15 +114,6 @@ class MapperFactory
         return $this->instances['MasterBrandMapper'];
     }
 
-    public function getPromotionMapper(): PromotionMapper
-    {
-        if (!isset($this->instances['PromotionMapper'])) {
-            $this->instances['PromotionMapper'] = new PromotionMapper($this);
-        }
-
-        return $this->instances['PromotionMapper'];
-    }
-
     public function getNetworkMapper(): NetworkMapper
     {
         if (!isset($this->instances['NetworkMapper'])) {
@@ -139,6 +130,15 @@ class MapperFactory
         }
 
         return $this->instances['OptionsMapper'];
+    }
+
+    public function getPromotionMapper(): PromotionMapper
+    {
+        if (!isset($this->instances['PromotionMapper'])) {
+            $this->instances['PromotionMapper'] = new PromotionMapper($this);
+        }
+
+        return $this->instances['PromotionMapper'];
     }
 
     public function getRelatedLinkMapper(): RelatedLinkMapper
--- a/src/Mapper/ProgrammesDbToDomain/MapperFactory.php
+++ b/src/Mapper/ProgrammesDbToDomain/MapperFactory.php
@@ -114,6 +114,15 @@ class MapperFactory
         return $this->instances['MasterBrandMapper'];
     }
 
+    public function getPromotionMapper(): PromotionMapper
+    {
+        if (!isset($this->instances['PromotionMapper'])) {
+            $this->instances['PromotionMapper'] = new PromotionMapper($this);
+        }
+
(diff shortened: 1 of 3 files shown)


src/Mapper/ProgrammesDbToDomain/MapperFactory.php
```diff
@@ -114,6 +114,15 @@ public function getMasterBrandMapper(): MasterBrandMapper
         return $this->instances['MasterBrandMapper'];
     }
 
+    public function getPromotionMapper(): PromotionMapper
```
BPScott: Alphabetical order please.

src/Mapper/ProgrammesDbToDomain/PromotionMapper.php
```diff
@@ -0,0 +1,87 @@
...
+                ->getDomainModel($dbPromotion['promotionOfCoreEntity']);
+        }
+
+        if (!empty($dbPromotion['promotionOfImage'])) {
+            return $this->mapperFactory
+                ->getImageMapper()
+                ->getDomainModel($dbPromotion['promotionOfImage']);
+        }
+
+        throw new InvalidArgumentException('A promotion has to promote something, but this one is promoting nothing');
```
BPScott: There is a difference between `$dbPromotion['blah']` being absent, and being null. Absent means "we didn't join to that table", while set but null means "we did join to that table but and there was nothing there".  I think we are incapable of judging if we have a valid Promotion if either of those keys do not exist, so you should check that and throw a `DataNotFetchedException`. See MasterBrandMapper->getNetwork() for an example of telling the difference. Currently you're assuming both of these keys exist.

You'll want two tests to confirm what happens when you don't pass in a valid coreEntity/image and when you don't join to either of those two tables

tests/Mapper/ProgrammesDbToDomain/PromotionMapperTest.php
```diff
@@ -0,0 +1,140 @@
...
+            'pid' => 'p000000h',
+            'startDate' => new DateTimeImmutable('1900-01-01 00:00:00.000000'),
+            'endDate' => new DateTimeImmutable('3000-01-01 00:00:00.000000'),
+            'weighting' => 73,
+            'isActive' => 1,
+            'promotedFor' => null,
+            'title' => 'active promotion of CoreEntity',
+            'uri' => 'www.myuri.com',
+            'cascadesToDescendants' => false,
+            'createdAt' => new DateTimeImmutable('2017-08-11 10:39:25.000000'),
```
BPScott: You can remove any values that aren't used to create the promotion object, as they're just adding noise.

tests/Mapper/ProgrammesDbToDomain/PromotionMapperTest.php
```diff
@@ -0,0 +1,140 @@
...
```
BPScott: You can remove any values that aren't used to create the promotion object, as they're just adding noise.

src/Mapper/ProgrammesDbToDomain/PromotionMapper.php
```diff
@@ -0,0 +1,87 @@
...
+            $this->getSynopses($dbPromotion),
+            $dbPromotion['uri'],
+            $dbPromotion['weighting']
+        );
+    }
+
+    /**
+     * @param array[] $dbPromotion
+     */
+    private function getSynopses(array $dbPromotion): Synopses
```
votemike: Does this crop up all over the place? Would it be worth turning in to a trait?

tatitati: I thought something similar in here. But I think we can make just another PR that move all this getSynopses() methods to a trait. Otherwise my PR would be noisy I think mixing different things.
It appear (repeated in these places):

PromotionMapper
CoreEntityMapper
ImageMapper
RelatedLinkMapper
SegmentEventMapper

PR #148, opened 2017-08-16
Added Promotions service

tatitati:


src/Service/PromotionsService.php
```diff
@@ -0,0 +1,82 @@
...
+        PromotionRepository $repository,
+        PromotionMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    /**
+     * @return Promotion[]
+     * @throws DataNotFetchedException if we get a promotion but we don't data about what is promoting
```
votemike: This description doesn't quite make sense.

src/Service/ServiceFactory.php
```diff
@@ -172,6 +172,19 @@ public function getProgrammesService(): ProgrammesService
         return $this->instances['ProgrammesService'];
     }
 
+    public function getPromotionsService(): PromotionsService
+    {
+        if (!isset($this->instances['PromotionsService'])) {
```
votemike: 'ProgrammesService' is sort of a magic string that keeps getting written. Should it be a variable?

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
```diff
@@ -0,0 +1,74 @@
...
+        $this->setUpCache();
+        $this->setUpRepo('PromotionRepository');
+        // override getDomainModel() in mapper with getDomainModelFromDbData()
+        $this->setUpMapper('PromotionMapper', 'getDomainModelFromDbData');
+    }
+
+    /**
+     * @param array[] $dbPromotion
+     */
+    protected function getDomainModelFromDbData($dbPromotion): Promotion
```
votemike: Typehint array?

src/Service/ServiceFactory.php
tatitati: I love constants for these cases, but I decided to follow they way is done so far. In the factory appears simple strings to define service names

votemike:
Looks OK to me.

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,84 @@
...
+        PromotionRepository $repository,
+        PromotionMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    /**
+     * @return Promotion[]
+     * @throws DataNotFetchedException if we get a promotion but there is nothing being
```
BPScott: This method does not directly throw an exception. `@throws` should only be placed functions that directly throw, and not on functions that call functions that throw. Otherwise basically every function shall end up decorated with an `@throws` which would be very noisy.

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,84 @@
...
+     *     promoted (an image or core entity) associated to the promotion
+     */
+    public function findActivePromotionsByProgramme(
+        Programme $programme,
+        DateTimeImmutable $dateTime,
```
BPScott: This param can be removed, and you can instead pass `ApplicationTime::getTime()` into the repository method.

We're never going to want to retrieve promos active on dates other than *right now*.

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,84 @@
...
+                );
+
+                return $this->mapManyEntities($dbEntities);
+            }
+        );
+    }
+
+    /**
+     * @return Promotion[]
+     * @throws DataNotFetchedException if we get a promotion but there is nothing being
```
BPScott: See above

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,84 @@
...
+     *     promoted (an image or core entity) associated to the promotion
+     */
+    public function findActiveSuperPromotionsByProgramme(
+        Programme $programme,
+        DateTimeImmutable $dateTime,
```
BPScott: See above

src/Service/ServiceFactory.php
BPScott: Swapping these stringy cache keys with constants sounds like a good idea, but let's do that all at once in a new PR. We could do it for both the ServiceFactory and the MapperFactory as they both follow a similar pattern.

How about using the name of the class you're caching an instance of? e.g.

```php
    public function getSegmentsService(): SegmentsService
    {
        if (!isset($this->instances[SegmentsService::class])) {
            $this->instances[SegmentsService::class] = new SegmentsService(
                $this->entityManager->getRepository('ProgrammesPagesService:Segment'),
                $this->mapperFactory->getSegmentMapper(),
                $this->cache
            );
        }

        return $this->instances[SegmentsService::class];
    }
```

tests/Service/PromotionsService/FindActivePromotionsByProgrammeTest.php
```diff
@@ -0,0 +1,29 @@
...
+class FindActivePromotionsByProgrammeTest extends AbstractPromotionServiceTest
+{
+    public function testServiceFindActivePromotions()
+    {
+        $mockProgramme = $this->getMockProgramme();
+        $dateTime = new DateTimeImmutable('1990-01-01 13:30:50');
+
+        $this->mockRepository
+            ->method('findActivePromotionsByPid')
+            ->with($mockProgramme->getPid(), $dateTime, 300, 0)
```
BPScott: This use of $mockProgramme has realised we should be passing a db ID into the repository function, this would allow us to save joining onto the core_entity table to find the DB id with that pid. Once that is done, $this->getMockProgramme can be replaced with `$this->getEntity('Brand', 1)`. Check FindByContributionToProgrammeTest for an example of doing this

tests/Service/PromotionsService/FindActiveSuperPromotionsByProgrammeTest.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Tests\BBC\ProgrammesPagesService\Service\PromotionsService;
+
+use DateTimeImmutable;
+
+class FindActiveSuperPromotionsByProgrammeTest extends AbstractPromotionServiceTest
```
BPScott: Test name should match the name of the function you're testing `FindActiveSuperPromotionsByAncestryTest`

src/Service/ServiceFactory.php
votemike: Sounds fair enough to me. Does someone want to create a ticket for that.

src/Service/PromotionsService.php
votemike: Why pass `ApplicationTime::getTime()` in to the repository method but not this method? Surely it should be both or neither?

src/Service/ServiceFactory.php
tatitati: I will

src/Service/PromotionsService.php
BPScott: The convention is that repository methods require no external state (such as a global application time that is not provided as part of the params). This allows them to be flexible so that we have the potential to reuse queries across multiple services methods. They are the core of our library so they should be functional (in the only use parameters, no side-effects sense).

Services meanwhile are a bit more fixed with their functionality, "getActivePromotions" is a common piece of functionality and we're never going to need to find promotions that were active at a point other than now. Services are the point where we encapsulate business logic, so they can have a bit more global state.

Thus I think this is the right layer to make use of ApplicationTime, rather than within the Controller (a layer up) or within the Repo (a later down).

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $pid = new Pid($dbPromotion['pid']);
+
+        $mockPromotion = $this->createMock(self::ENTITY_NS . 'Promotion');
+        $mockPromotion->method('getPid')->willReturn($pid);
+        return $mockPromotion;
+    }
+
+    /**
+     * @param array[] $dbPromotions
```
votemike: Could this also be `mixed[][]`?

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return Promotion[]
+     */
+    protected function getDomainModelsFromDbData(array $dbPromotions): array
+    {
+        return array_map([$this, 'getDomainModelFromDbData'], $dbPromotions);
+    }
+
+    protected function getMockProgramme(): Brand
+    {
+        $coreEntity = $this->createMock('BBC\ProgrammesPagesService\Domain\Entity\Brand');
```
votemike: `Brand::class` instead?

votemike: Is `$coreEntity` the best name for this? The method is called `getMockProgramme()`, you're mocking a `Brand` and calling it `$coreEntity`.  
What is it you're actually testing? Is it something that implements a `PromotableInterface`? If so, maybe rename some of these things to make it more obvious it's a `PromotableInterface` you're testing.

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,77 @@
...
+                return $this->mapManyEntities($dbEntities);
+            }
+        );
+    }
+
+    /**
+     * @return Promotion[]
+     */
+    public function findActiveSuperPromotionsByEntity(
+        CoreEntity $coreEntity,
```
votemike: Should this be `PromotableInterface` instead of CoreEntity?

tatitati: no. Actually is a little confuse. Maybe a better name should be ContextEntity. The promotable one is the thing that is being promoted. The context is "who" is promoting it. Context is maybe better

BPScott: Consider the idea "I'm on the Doctor Who brand page, give me all promotions relating to doctor who". In this case we pass in the Doctor Who Brand as the context. We get back a bunch of Promotions that are promoting either CoreEntities or Images (which are encapsulated as PromotableInterface).

tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActivePromotionsByContextTest.php
```diff
@@ -1,45 +0,0 @@
-<?php
```
BPScott: Removing this should be in the other PR

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,72 @@
...
+        PromotionMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    /**
+     * @return Promotion[]
+     */
+    public function findActivePromotionsByEntity(
```
BPScott: findActivePromotionsByContext?

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,72 @@
...
+                );
+
+                $promotions = $this->mapManyEntities($dbEntities);
+                return $this->groupPromotionsByType($promotions);
+            }
+        );
+    }
+
+    /**
+     * @param Promotion[] $promotions
```
votemike: /**
 * @param Promotion[] $promotions
 * @return Promotion[][]
*/

tatitati: arg....I really hate this limitation on PHP, is still so little descriptive.....
I remember in some of my previous companies we were using swagger to be more descriptives. 
https://apihandyman.io/writing-openapi-swagger-specification-tutorial-part-4-advanced-data-modeling/#arrays-size-and-uniqueness

I know is more verbose, and is not a PHP standard, but swagger is also wide used and is supported by many tools. I'm dropping this thought because some notation is still quite confuse

votemike: Interesting. Maybe make my change for now. But bring up Swagger in a meeting. Maybe retro?

tatitati: The last one I used was blueprints, but in my opinion I prefer swagger (the syntax is quite similar to YML which is very clean). Blue prints has many symbols

tests/Service/PromotionsService/FindActivePromotionsByEntityTest.php
```diff
@@ -0,0 +1,70 @@
...
+     * @var CoreEntity
+     */
+    private $context;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $dbId = 1;
+        $context = $this->createMock(Brand::class, $dbId);
```
BPScott: You can use a utility function:`$this->mockEntity('Brand', 1)`

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,74 @@
...
+                    $limit,
+                    $this->getOffset($limit, $page)
+                );
+
+                return $this->mapManyEntities($dbEntities);
+            }
+        );
+    }
+
+    public function findActivePromotionsByEntityGropuedByType(
```
BPScott: "Grouped" not "Gropued"

tatitati: ouch!

src/Service/PromotionsService.php
```diff
@@ -0,0 +1,74 @@
...
+                    $limit,
+                    $this->getOffset($limit, $page)
+                );
+
+                return $this->mapManyEntities($dbEntities);
+            }
+        );
+    }
+
+    public function findActivePromotionsByEntityGroupedByType(
```
votemike:
```
/**
   * @return Promotion[][]
   */
```

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $dbData = [['pid' => 'p000000h']];
+
+        $this->mockRepository
+            ->method('findActivePromotionsByContext')
+            ->with($this->context->getDbAncestryIds(), $this->isInstanceOf(DateTimeImmutable::class), 300, 0)
+            ->willReturn($dbData);
+
+        $promotions = $this->service()->findActivePromotionsByContext($this->context);
+
+        $this->assertEquals($this->getDomainModelsFromDbData($dbData), $promotions);
```
votemike: Is this assertion actually useful? The expects is fake data, as is the actual. No?

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $this->setUpCache();
+        $this->setUpRepo('PromotionRepository');
+        // override getDomainModel() in mapper with getDomainModelFromDbData()
+        $this->setUpMapper('PromotionMapper', 'getDomainModelFromDbData');
+    }
+
+    /**
+     * @param mixed[] $dbPromotion
+     */
+    protected function getDomainModelFromDbData(array $dbPromotion): Promotion
```
votemike: I'm suspicious of this method. It doesn't seem to be something that should be needed in a test.

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $dbData = [['pid' => 'p000000h']];
+
+        $this->mockRepository
+            ->method('findActivePromotionsByContext')
+            ->with($this->context->getDbAncestryIds(), $this->isInstanceOf(DateTimeImmutable::class), $expectedLimit, $expectedOffset)
+            ->willReturn($dbData);
+
+        $promotions = $this->service()->findActivePromotionsByContext($this->context, 30, 3);
+
+        $this->assertEquals($this->getDomainModelsFromDbData($dbData), $promotions);
```
votemike: Again, not convinced this is a useful assertion.

tests/Service/PromotionsService/FindActivePromotionsByEntityGroupedByTypeTest.php
```diff
@@ -0,0 +1,43 @@
...
+        parent::setUp();
+
+        $dbId = 1;
+        $context = $this->mockEntity('Brand', $dbId);
+        $context->method('getDbAncestryIds')->willReturn([$dbId]);
+
+        $this->context = $context;
+    }
+
+    public function testfindActivePromotionsByEntityGroupedByTypeReturnRightStructure()
```
votemike: How does this test prove that all promotions aren't just put in the super category?

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    /** @var CoreEntity */
+    private $context;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $dbId = 1;
+        $context = $this->mockEntity('Brand', $dbId);
```
votemike: Can you just mock 'CoreEntity'?

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace Tests\BBC\ProgrammesPagesService\Service\PromotionsService;
+
+use BBC\ProgrammesPagesService\Domain\Entity\Brand;
```
votemike: Is this used? You could run [Optimize Imports](https://www.jetbrains.com/help/idea/optimizing-imports.html) (`⌥⌘O`) to remove any that you're not using instead of going through 1 by 1.

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
tatitati: Hi Michael, to be honest I think something similar to you. 

In this case we are mocking the db data, and the domain model Promotion is later not created with the mapper and mocked again. The only tests in here that make sense for me should be"I expect this method to be called, and this one as well". But compare data in here doesn't make sense. 

I did this following what we were making so far, and I think this should be improved in any way. Thinking about it...... Thanks for the comment, I thought was only me 😄

votemike: Exactly.  
We expect the cache methods `keyHelper()` and `getOrSet()` are called.  
We expect the repositories `findActivePromotionsByContext()` method is called with _X_ arguments and returns _Y_.  
And we expect the mappers `getDomainModel()` method is called _Z_ times.

tests/Service/PromotionsService/FindActivePromotionsByEntityGroupedByTypeTest.php
tatitati: Just updated (sorry, I was late) 😄

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
BPScott: It's fake data yes, but it proves that the Service returns what came out of the Mapper (with any additional mutations if needed, in this case there no additional mutations are needed)

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
BPScott: It's a tool for mocking out the Mapper, otherwise the service tests would end up testing the Service and Mapper functionality.

This is the style that we use for all Service tests so if we change it, we should do it all at once rather than in this PR.

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
votemike: It sounds like this test is testing the Mapper then.... which doesn't sound right.

BPScott: Those tests are all for mocking inside the function. This line tests that the function returns a particular value, which is not currently done elsewhere

tests/Service/PromotionsService/AbstractPromotionServiceTest.php
votemike: OK then.

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
votemike: Would asserting some other things on the returned value be useful? Such as the number of items returned. Or the value of a specific item rather than just the whole return value equals this whole value?

tests/Service/PromotionsService/FindActivePromotionsByEntityGroupedByTypeTest.php
```diff
@@ -0,0 +1,54 @@
...
+        ];
+
+        $this->mockRepository
+            ->method('findActivePromotionsByContext')
+            ->with($this->context->getDbAncestryIds(), $this->isInstanceOf(DateTimeImmutable::class), 300, 0)
+            ->willReturn($dbData);
+
+        $promotions = $this->service()->findActivePromotionsByEntityGroupedByType($this->context);
+
+        $this->assertEquals(
```
votemike: Would it be better to split this up into several assertions? Asserting array lengths, asserting the array has the keys regular and super etc....?

tatitati: well, I know that the ideal would be an assert per test, but is also quite usual to find multiple asserts in only one test (if they are related)

tests/Service/PromotionsService/FindActivePromotionsByContextTest.php
tatitati: Thanks ben for the explanation

tests/Service/PromotionsService/FindActivePromotionsByEntityGroupedByTypeTest.php
votemike: It's fine to have more than one assert per test. I mean, should this be split up in to multiple assertions?

tatitati: well, we can create multiple asserts (for length, type, keys, etc). But also we can assert all that in once. I think is more readable in this way, because you show exactly what you expect in once

tests/Service/PromotionsService/FindActivePromotionsByEntityGroupedByTypeTest.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    /** @var CoreEntity */
+    private $context;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $dbId = 1;
+        $context = $this->mockEntity('Brand', $dbId);
```
votemike: Could this be 'CoreEntity'?

PR #152, opened 2017-08-16
GetSynopses() moved to trait

tatitati:


src/Mapper/Traits/SynopsesTrait.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Mapper\Traits;
+
+use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+
+trait SynopsesTrait
+{
+    private function getSynopses(array $dbImage): Synopses
```
votemike: https://gist.github.com/votemike/04d7f5c2859a6999a5aa3915ea40c982#file-arraydocblocks-php-L55

src/Mapper/ProgrammesDbToDomain/SegmentMapper.php
```diff
@@ -5,12 +5,14 @@
...
 use BBC\ProgrammesPagesService\Domain\Entity\Segment;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
-use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+use BBC\ProgrammesPagesService\Mapper\Traits\SynopsesTrait;
 
 class SegmentMapper extends AbstractMapper
 {
     const MUSIC_TYPES = ['music', 'classical'];
 
+    use SynopsesTrait;
```
BPScott: Should `use` blocks be before `const` declarations?

src/Mapper/Traits/SynopsesTrait.php
```diff
@@ -0,0 +1,20 @@
...
+namespace BBC\ProgrammesPagesService\Mapper\Traits;
+
+use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+
+trait SynopsesTrait
+{
+    /**
+     * @param mixed[] $dbImage
+     */
+    private function getSynopses(array $dbImage): Synopses
```
BPScott: `$dbEntity` as it won't always be an image?

votemike:
Does the trait need a test?

votemike:
Looking at the deleted code, the method wasn't always the same, some of them were missing typehints. Good we refactored 👍

PR #153, opened 2017-08-17
Promotions: Simplified query and added field in domain model

tatitati:
We could fetch 'all active promotions' and 'only active superpromotions'. But we saw that in same cases would be useful to fetch 'only active No-superpromotions'


tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActiveSuperpromotionsByContextTest.php
```diff
@@ -9,7 +9,7 @@
 /**
```
tatitati: Sorry for the noise in tests. Tests are the same, is just that I reordered, and change some numbers (according to the new order I was following )

src/Domain/Entity/Promotion.php
```diff
@@ -7,6 +7,9 @@
 
 class Promotion
 {
+    /** @var  bool */
+    private $isSuperPromotion;
```
BPScott: I find it easier to scan if the properties are in the same order as they were passed into the constructor

src/Domain/Entity/Promotion.php
```diff
@@ -31,14 +34,21 @@ public function __construct(
...
         $this->pid = $pid;
         $this->promotedEntity = $promotedEntity;
         $this->synopses = $synopses;
         $this->title = $title;
         $this->url = $url;
         $this->weighting = $weighting;
+        $this->isSuperPromotion = $isSuperPromotion;
+    }
+
+    public function getIsSuperPromotion(): bool
```
BPScott: isSuperPromotion()

Should we stick this at the bottom so the getters are in the same order as the constructor args? That's what all the other entities do.

tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActivePromotionsByContextTest.php
```diff
@@ -10,25 +11,95 @@
...
+        $this->assertCount(3, $this->getDbQueries());
+    }
 
-        $this->assertEquals(['p000000h', 'p000001h', 'p000004h'], array_column($dbPromotions, 'pid'));
+    public function testActiveSuperpromotionsAreReceivedWithSuperPromotionsForSeries()
+    {
+        $coreEntityRepo = $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity');
+
+        $series = $coreEntityRepo->findByPid('b00swyx1', 'Series'); // series 1/episode 1
+        $seriesDbAncestryIds = array_filter(explode(',', $series['ancestry']));
```
BPScott: `$this->getAncestryFromPersistentIdentifier('b00swyx1', 'CoreEntity')` is a utility method provided by the base class to do this. It's particularly useful because it doesn't count the query when you call `getDbQueries`.

tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActivePromotionsByContextTest.php
```diff
@@ -10,25 +11,95 @@
...
 
+        $this->assertCount(3, $this->getDbQueries());
+    }
+
+    public function testActiveSuperpromotionsAreReceivedWithSuperPromotionsForBrand()
+    {
+        $coreEntityRepo = $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity');
+
+        $brand = $coreEntityRepo->findByPid('b010t19z', 'Brand');
+        $brandDbAncestryIds = array_filter(explode(',', $brand['ancestry']));
```
BPScott: $this->getAncestryFromPersistentIdentifier('b010t19z', 'CoreEntity') is a utility method provided by the base class to do this. It's particularly useful because it doesn't count the query when you call getDbQueries.

tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActivePromotionsByContextTest.php
```diff
@@ -10,25 +11,95 @@
...
+
+        $this->assertCount(3, $this->getDbQueries());
+    }
+
+    public function testActiveSuperpromotionsAreReceivedWithSuperPromotionsForImage()
+    {
+        $coreEntityRepo = $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity');
+
+        $brand = $coreEntityRepo->findByPid('b010t150', 'Series'); // brand 1/series 2
+        $brandDbAncestryIds = array_filter(explode(',', $brand['ancestry']));
```
BPScott: $this->getAncestryFromPersistentIdentifier('b010t150', 'CoreEntity') is a utility method provided by the base class to do this. It's particularly useful because it doesn't count the query when you call getDbQueries.

tests/Data/ProgrammesDb/EntityRepository/PromotionRepository/FindActivePromotionsByContextTest.php
```diff
@@ -10,25 +11,95 @@
  */
 class FindActivePromotionsByContextTest extends AbstractDatabaseTest
 {
-    public function testActivePromotionsAreReceived()
+    /**
+     * @var PromotionRepository
```
BPScott: Use the short docblock style to save on line-noise

```php
/** @var PromotionRepository /*
```

tests/DataFixtures/ORM/PromotionsFixture.php
```diff
@@ -22,57 +24,102 @@ public function getDependencies()
...
     {
+        /*
+        PROMOTION CONTEXTS:
+        ===================
+
+        Brand (b010t19z)
+            promotion 0 (p000000h, SUPER PROMOTION)
+
+            series 1 (b00swyx1)
+                promotion 1 (p000001h)
```
BPScott: Would it be worth having a super promo on the series or episode to prove that the inheritance is working correctly?

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
```diff
@@ -13,8 +13,12 @@ class PromotionRepository extends EntityRepository
     /**
      * @return array[]
      */
-    public function findActivePromotionsByContext(int $contextId, DateTimeImmutable $datetime, ?int $limit, int $offset): array
+    public function findActivePromotionsByContext(array $ancestryIds, DateTimeImmutable $datetime, ?int $limit, int $offset): array
```
votemike: Docblock for $ancestryIds

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
```diff
@@ -23,38 +27,19 @@ public function findActivePromotionsByContext(int $contextId, DateTimeImmutable
...
-           ->andWhere('promotion.startDate <= :datetime')
-           ->andWhere('promotion.endDate > :datetime')
-           ->andWhere('promotion.context in (:ancestryIds)')
-           ->andWhere('promotion.cascadesToDescendants = 1')
-           ->addOrderBy('promotion.weighting', 'ASC')
-           ->setFirstResult($offset)
-           ->setMaxResults($limit)
-           ->setParameter('ancestryIds', $ancestryIds)
-           ->setParameter('datetime', $datetime);
+        if (!$ancestryIds) {
```
votemike: Could this check be reversed?

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
```diff
@@ -13,8 +13,12 @@ class PromotionRepository extends EntityRepository
     /**
      * @return array[]
      */
-    public function findActivePromotionsByContext(int $contextId, DateTimeImmutable $datetime, ?int $limit, int $offset): array
+    public function findActivePromotionsByContext(array $ancestryIds, DateTimeImmutable $datetime, ?int $limit, int $offset): array
     {
+        // $contextId is the current context. We want to get both cascading and non-cascading promos for this id
+        // $ancestryIds is an array ids of the parents of the current context. For these ids we only want cascading promos
+        $contextId = array_pop($ancestryIds);
```
votemike: What happens if $ancestryIds is an empty array? Do we get a sensible error?

tests/DataFixtures/ORM/PromotionsFixture.php
```diff
@@ -22,57 +24,102 @@ public function getDependencies()
...
-            'active promotion of CoreEntity',
-            $this->getReference('b00swyx1'), // promotion of Core entity - series 1
-            1, // weight
+            'promotion 1',
+            $this->getReference('b0175lqm'), // promoting: series2/episode1
+            5,
             true,
             new DateTime('1900-01-01 00:00:00'),
-            new DateTime('3000-01-01 00:00:00')
+            new DateTime('3000-01-01 00:00:00'),
```
votemike: This test will start failing in 983 years 😝

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
tatitati: There is not core_entities with null ancestry values = null. 
Also in the limit case, the coreentity_id (the context_id in here) is the same than the ancestry.

id |   PID        |           Ancestry
11|	b006q2x0|	11,

tests/DataFixtures/ORM/PromotionsFixture.php
tatitati: impossible!, my code is designed to work one year more! (989)

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
votemike: What if someone is manually building the array to send through and mucks it up so that it ends up empty?

tatitati: ok, to avoid a wrong use of the function, I see your point, Thanks!

tests/DataFixtures/ORM/PromotionsFixture.php
BPScott: [Thanks to PHP having weirdly niche but occasionally neat and expressive stuff in it](http://php.net/manual/en/datetime.formats.relative.php) you could do `new DateTime('-1 year')` and `new DateTime('+1 year')` to generate relative timestamps?

Not convinced that's more readable than fixed far-future things though :)

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
BPScott: Good call, an explanatory comment on what this is doing and why would be handy too.

PR #157, opened 2017-08-22
[Tests tidy] Broadcasts service 

tatitati:
Main changes:
- Added extra tests to test pagination on FindByVersionTest.php
- Removed unused function broadcastsFromDbData() (is not used anymore in order to get more visibility on tests)
- Modified explicit dates for dynamics "+/-1year"
- Simplified some tests

### DIFF
--- a/tests/AbstractServiceTest.php
+++ b/tests/AbstractServiceTest.php
@@ -40,7 +40,7 @@ abstract class AbstractServiceTest extends TestCase
     {
         $this->mockMapper = $this->createMock($this::MAPPER_NS . $mapperName);
 
-        $this->mockMapper->expects($this->any())
+        $this->mockMapper
             ->method('getDomainModel')
             ->will($this->returnCallback(function ($entity) use ($entityBuilderMethod) {
                 return call_user_func([$this, $entityBuilderMethod], $entity);
--- a/tests/AbstractServiceTest.php
+++ b/tests/AbstractServiceTest.php
@@ -36,15 +36,18 @@ abstract class AbstractServiceTest extends TestCase
             ->getMock();
     }
 
-    protected function setUpMapper($mapperName, $entityBuilderMethod)
+    protected function setUpMapper($mapperName, $callback)
     {
         $this->mockMapper = $this->createMock($this::MAPPER_NS . $mapperName);
 
-        $this->mockMapper
-            ->method('getDomainModel')
-            ->will($this->returnCallback(function ($entity) use ($entityBuilderMethod) {
-                return call_user_func([$this, $entityBuilderMethod], $entity);
-            }));
+        // TODO once we always pass in a callable to this function then add a typehint and remove this check
+        if (!is_callable($callback)) {
+            $callback = function ($entity) use ($callback) {
+                return $this->$callback($entity);
+            };
+        }
+
+        $this->mockMapper->method('getDomainModel')->will($this->returnCallback($callback));
     }
 
     protected function mockEntity($name, $dbId = null)
@@ -56,5 +59,19 @@ abstract class AbstractServiceTest extends TestCase
         }
 
         return $entity;
+    }
+
+    /**
+     * @param array $entities
+     * @return string[]
+     */
+    protected function extractPids(array $entities): array
+    {
+        return array_map(
+            function ($entity) {
+                return (string) $entity->getPid();
+            },
+            $entities
+        );
     }
 }
--- a/tests/AbstractServiceTest.php
+++ b/tests/AbstractServiceTest.php
@@ -62,7 +62,7 @@ abstract class AbstractServiceTest extends TestCase
     }
 
     /**
-     * @param array $entities
+     * @param array $entities any model domain with getPid() function
      * @return string[]
      */
     protected function extractPids(array $entities): array
--- a/tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
+++ b/tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
@@ -9,7 +9,7 @@ use Tests\BBC\ProgrammesPagesService\AbstractServiceTest;
 
 abstract class AbstractBroadcastsServiceTest extends AbstractServiceTest
 {
-    public function setUp(): void
+    public function setUp()
     {
         $this->setUpCache();
         $this->setUpRepo('BroadcastRepository');
--- a/tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
+++ b/tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
@@ -13,13 +13,13 @@ abstract class AbstractBroadcastsServiceTest extends AbstractServiceTest
     {
         $this->setUpCache();
         $this->setUpRepo('BroadcastRepository');
-        $this->setUpMapper('BroadcastMapper', 'fakeGetDomainModel');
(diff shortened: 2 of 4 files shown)


tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
```diff
@@ -2,31 +2,28 @@
...
-    public function setUp()
+    public function setUp(): void
     {
         $this->setUpCache();
         $this->setUpRepo('BroadcastRepository');
         $this->setUpMapper('BroadcastMapper', 'broadcastFromDbData');
     }
 
-    protected function broadcastsFromDbData(array $entities)
+    protected function broadcastFromDbData(array $dbData): Broadcast
```
votemike: Docblock for array?

votemike: Also `@return Broadcast|PHPUnit_Mock` or whatever the mock is.

tests/AbstractServiceTest.php
```diff
@@ -40,7 +40,7 @@ protected function setUpMapper($mapperName, $entityBuilderMethod)
     {
         $this->mockMapper = $this->createMock($this::MAPPER_NS . $mapperName);
 
-        $this->mockMapper->expects($this->any())
+        $this->mockMapper
```
votemike: Do we want this to be `->expects(1)`?

tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
```diff
@@ -2,31 +2,31 @@
...
+    public function setUp(): void
     {
         $this->setUpCache();
         $this->setUpRepo('BroadcastRepository');
         $this->setUpMapper('BroadcastMapper', 'broadcastFromDbData');
     }
 
-    protected function broadcastsFromDbData(array $entities)
+    /**
+     * @param array[] $dbData
```
votemike: This isn't an array of arrays is it? Shouldn't it be `mixed[]`?

tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
```diff
@@ -2,31 +2,31 @@
...
 {
-    public function setUp()
+    public function setUp(): void
     {
         $this->setUpCache();
         $this->setUpRepo('BroadcastRepository');
         $this->setUpMapper('BroadcastMapper', 'broadcastFromDbData');
     }
 
-    protected function broadcastsFromDbData(array $entities)
```
votemike: What about removing these types of methods from the other service tests?

tests/Service/BroadcastsService/FindByVersionTest.php
```diff
@@ -2,49 +2,65 @@
...
+        $version = $this->mockEntity('Version', $dbId);
+        $dbData = [['pid' => 'b00swyx1'], ['pid' => 'b010t150']];
+
+        $this->mockRepository
+            ->method('findByVersion')
+            ->willReturn($dbData);
+
+        $broadcasts = $this->service()->findByVersion($version);
+
+        $this->assertCount(2, $broadcasts);
```
votemike: 👍

tests/Service/BroadcastsService/FindByServiceAndDateRangeTest.php
```diff
@@ -2,47 +2,66 @@
...
+        $broadcasts = $this->service()->findByServiceAndDateRange($sid, $fromDateTime, $toDatetime);
 
-        $this->assertEquals($this->broadcastsFromDbData($dbData), $broadcasts);
+        $this->assertCount(2, $broadcasts);
+        $this->assertContainsOnly(Broadcast::class, $broadcasts);
+        $this->assertEquals('b00swyx1', (string) $broadcasts[0]->getPid());
+        $this->assertEquals('b010t150', (string) $broadcasts[1]->getPid());
     }
 
     public function testFindByServiceAndDateRangeWhenNoBroadcastsFound()
```
votemike: I feel this test is simple/similar enough to put in with the pagination provider. (Maybe change the method name of paginationProvider). It just avoids repetition.

tests/AbstractServiceTest.php
tatitati: Just checked. is not consistent the amount of calls to getDomainModel. So Expects($this->once()) breaks many tests. jeje. I will manage anyway...maybe I can do something as I go with the tidy up.

tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
tatitati: well, actually is an array of arrays. arrays represent here "rows" in db, and each row is a mixed array with the fields/values

tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
tatitati: On my way, but I will go service by service. :smile:

tests/Service/BroadcastsService/FindByServiceAndDateRangeTest.php
tatitati: arg.......I still think that this tests is testing something different of pagination. If this test fails one day, we will know quickly that pagination was working properly, and is related with this specific behaviour

tests/Service/BroadcastsService/AbstractBroadcastsServiceTest.php
votemike: If it's an array of arrays, how can you access `$dbData['pid']`? Is the value of `$dbData['pid']` an array?

tatitati: true, didn't spot it. That function is called per row, so is the mixed[]

tests/Service/BroadcastsService/FindByServiceAndDateRangeTest.php
```diff
@@ -2,48 +2,88 @@
...
+            ],
+        ];
+    }
 
-        $this->assertSame([], $broadcasts);
+    /**
+     * @param Broadcast[] $broadcasts
+     * @return string[]
+     */
+    private function extractPids(array $broadcasts): array
```
BPScott: We're going to use this in pretty much all services as they follow a similar pattern right? Would it be better placed in the AbstractServiceTest so everything can use it

tatitati: :+1:

tests/Service/BroadcastsService/FindByServiceAndDateRangeTest.php
```diff
@@ -37,7 +37,7 @@ public function paginationProvider()
             'custom pagination' => [3, 12, [3, 5]],
         ];
     }
-
+
```
votemike: Why?

tests/AbstractServiceTest.php
```diff
@@ -36,15 +36,18 @@ protected function setUpCache()
             ->getMock();
     }
 
-    protected function setUpMapper($mapperName, $entityBuilderMethod)
+    protected function setUpMapper($mapperName, $callback)
```
votemike: Typehint string and callable?

BPScott: The seccond arg is currently callable|string as we haven't retrofitted the rest of the service tests yet.

PR #161, opened 2017-08-24
mocking less verbose

tatitati:
I really would like to see this way of mocking more. Is simpler, and less verbose.

Why!?
With createMock() you need at least two lines, to mock a method. 

    $stubBroadcast = $this->createMock(Broadcast::class);
    $stubBroadcast->method('getPid')->willReturn(new Pid($dbData['pid']));

With createConfiguredMock():

    $this->createConfiguredMock(AtozTitle::class,['getFirstletter' => $dbData['firstLetter']])

Even if you format it is still more compact (no methods around, no arrows...):

    $this->createConfiguredMock(
                AtozTitle::class,
                ['getFirstletter' => $dbData['firstLetter']]
    );


BPScott:
Nice, I had no idea `createConfiguredMock` was a thing!

PR #163, opened 2017-08-30
Removing findBySids method

votemike:
Can't see this being used any where.


BjornTwachtmann:
Doesn't look like it's in use, should be fine to remove

PR #164, opened 2017-08-31
PhpStan

votemike:
Part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5778

Should also resolve https://github.com/bbc/programmes-frontend/issues/143

### DIFF
--- a/src/Domain/Entity/Service.php
+++ b/src/Domain/Entity/Service.php
@@ -163,10 +163,9 @@ class Service
     public function isTv(): bool
     {
         // Network can be null, but unfetched should throw an exception
-        if ($this->network && $this->getNetwork()->isTv()) {
-            return true;
-        }
-        return false;
+        $network = $this->getNetwork();
+
+        return $network && $network->isTv();
     }
 
     /**
@@ -175,17 +174,15 @@ class Service
     public function isRadio(): bool
     {
         // Network can be null, but unfetched should throw an exception
-        if ($this->network && $this->getNetwork()->isRadio()) {
-            return true;
-        }
-        return false;
+        $network = $this->getNetwork();
+
+        return $network && $network->isRadio();
     }
 
     public function isInternational(): bool
     {
-        if ($this->network && $this->getNetwork()->isInternational()) {
-            return true;
-        }
-        return false;
+        $network = $this->getNetwork();
+
+        return $network && $network->isInternational();
     }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -2155,16 +2155,16 @@
         },
         {
             "name": "phpstan/phpstan",
-            "version": "0.8.3",
+            "version": "0.8.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/phpstan/phpstan.git",
-                "reference": "ccc046ca473dcdb5ce9225cc05d7808f2e327f40"
+                "reference": "7745ab983ff05e0c4ca2c94e26c113ab734db3ce"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/phpstan/phpstan/zipball/ccc046ca473dcdb5ce9225cc05d7808f2e327f40",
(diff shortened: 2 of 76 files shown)


src/Domain/Entity/Service.php
```diff
@@ -175,17 +174,15 @@ public function isTv(): bool
     public function isRadio(): bool
     {
         // Network can be null, but unfetched should throw an exception
-        if ($this->network && $this->getNetwork()->isRadio()) {
-            return true;
-        }
-        return false;
+        $network = $this->getNetwork();
+
+        return $network && $$network->isRadio();
```
BPScott: $$network?

src/Domain/Entity/Service.php
```diff
@@ -175,17 +174,15 @@ public function isTv(): bool
...
 
     public function isInternational(): bool
     {
-        if ($this->network && $this->getNetwork()->isInternational()) {
-            return true;
-        }
-        return false;
+        $network = $this->getNetwork();
+
+        return $network && $$network->isInternational();
```
BPScott: $$network?

BjornTwachtmann:
I'm pretty unconvinced of the utility of casting everything to a string before passing it into the keyHelper, if we're not sure they're strings then remove the typehint from keyHelper and cast them in that function. This is a loooot of boilerplate. 

Other than that, I'm fine with these changes in theory, but before merging I think we should do a Faucet branch with these changes in and leave it deployed to INT for a week or so to make sure it doesn't cause any issues. Faucet can be touchy.

src/Data/ProgrammesDb/Entity/EntityInterfaces/RelatedInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces;
+
+interface RelatedInterface
```
BPScott: Pondering on some better names for these interfaces: RelatedLinkContextInterface?

src/Data/ProgrammesDb/Entity/EntityInterfaces/MemberInterface.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces;
+
+interface MemberInterface
```
BPScott: Pondering on some better names for these interfaces: MemberOfGroupInterface?

src/Domain/Entity/Service.php
BjornTwachtmann: Hmm. Looking at this (aside from the typo), this will throw an exception if network is unfetched, where the old code would have simply returned false. Is that what we want?

votemike: The old code would also have thrown the exception. Wouldn't it?

BPScott: I think we should be clear and shouty when we try and access some information that we haven't retrieved yet to avoid those insidious "why is everything powering through and returning false" bugs.

BPScott:
Good idea with the casting within keyHelper, that'll help reduce a bunch of noise within the services :)

votemike:
test this please

votemike:
@BjornTwachtmann before merging this branch in to master you want the faucet branch deployed? Why would Faucet break after these changes? That suggests that the tests for this package aren't good enough...

felipeparaujo:
@votemike These changes shouldn't break Faucet, but Faucet is very finicky. A good way to stress test these changes is to let Faucet do its thing for a while.

votemike:
@BjornTwachtmann @felipeparaujo OK. Here's the deploy https://cosmos.tools.bbc.co.uk/deployments/1154045

BjornTwachtmann:
Looks fine now. We'll merge this after a few days once we've verified Faucet is stable.

PR #165, opened 2017-09-01
Add getShortSynopsis to Promotion

BPScott:
Its the Synopsis we pretty much always want.

Fix a tiny spacing issue in a comment while I'm in there

### DIFF
--- a/src/Domain/Entity/Promotion.php
+++ b/src/Domain/Entity/Promotion.php
@@ -25,7 +25,7 @@ class Promotion
     /** @var int */
     private $weighting;
 
-    /** @var  bool */
+    /** @var bool */
     private $isSuperPromotion;
 
     public function __construct(
@@ -59,6 +59,11 @@ class Promotion
     public function getSynopses(): Synopses
     {
         return $this->synopses;
+    }
+
+    public function getShortSynopsis(): string
+    {
+        return $this->synopses->getShortSynopsis();
     }
 
     public function getTitle(): string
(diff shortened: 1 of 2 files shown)


src/Domain/Entity/Promotion.php
```diff
@@ -25,7 +25,7 @@ class Promotion
     /** @var int */
     private $weighting;
 
-    /** @var  bool */
+    /** @var bool */
```
votemike: Single line comments also only have 1 leading start don't they?

src/Domain/Entity/Promotion.php
```diff
@@ -61,6 +61,11 @@ public function getSynopses(): Synopses
         return $this->synopses;
     }
 
+    public function getShortSynopsis(): string
```
votemike: If a promotion already has a`getSynopses()` method, why do we need this one?

src/Domain/Entity/Promotion.php
BPScott: It's a docblock rather than a general comment so it has [two asterisks](https://docs.phpdoc.org/guides/docblocks.html). If you don't have two asterisks then phpdoc won't pick it the annotations within the comment.

src/Domain/Entity/Promotion.php
BPScott: Utility - it all cases i can think of so far we want the short Synopsis and dont need the others. To save a bit of indirection it's a bit neater to do this (same as things like $service->isTV() are aliases to $service->getNetwork()->isTv()). CoreEntity also provides a getSynopsis and a getShortSynopsis

votemike: But the `isTv()` utility is there because Network might be null....  
I feel it's unnecessary. But the code looks fine if you still feel it's useful.

PR #166, opened 2017-09-01
Attach RelatedLinks to Promotions

BPScott:
Whenever we display promotions we always want to show the related links
attached to them. As there are very few promos with more than one related
link it is cheaper to request them in the same query as the promos, rather
than do a two step hydration process.

* Add a OneToMany 'relatedLinks' relationship into the Promotion Doctrine
  entity (this is the inverse of the relationship on the RelatedLink).
  This does not require a DB change, it is just a mapping update.
* Add an array of related links to the Promotion's constructor, this may
  be null to signify the Related Links were unfetched (this is the same
  as how Genres/Formats on a Programme works)
* Update the PromotionMapper to construct Promotions with Related Links
  if they were requested.
* Request Related Links in
  PromotionRepository->findActivePromotionsByContext()
* Update the PromotionRepository test to demo requesting related links

### DIFF
--- a/src/Data/ProgrammesDb/Entity/Promotion.php
+++ b/src/Data/ProgrammesDb/Entity/Promotion.php
@@ -120,6 +120,14 @@ class Promotion implements RelatedLinkContextInterface
      */
     private $cascadesToDescendants = false;
 
+    /**
+     * Used in queries only. No getter or setter as creation should be handled
+     * by the RelatedLink entity.
+     *
+     * @ORM\OneToMany(targetEntity="RelatedLink", mappedBy="relatedToPromotion")
+     */
+    private $relatedLinks;
+
     public function __construct(
         string $pid,
         PromotableInterface $promotionOf,
--- a/src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
@@ -27,8 +27,18 @@ class PromotionRepository extends EntityRepository
         $contextId = end($ancestryIds);
         $parentIds = array_slice($ancestryIds, 0, -1);
 
+        // YIKES! relatedLinks is a many-to-many join, that could result in an
+        // increased number of rows returned by the DB and the potential for
+        // slow DB queries as per https://ocramius.github.io/blog/doctrine-orm-optimization-hydration/.
+        // Except it doesn't - the vast majority of Promotions have zero or one
+        // RelatedLink. At time of writing this comment (Sept 2017) only 2.85%
+        // of the Promotions in PIPS have 2 or more RelatedLinks and 0.10% have
+        // 5 or more RelatedLinks. Creating an few extra rows in rare cases is
+        // way more efficient that having to do a two-step hydration process.
+
         $qb = $this->createQueryBuilder('promotion')
-            ->addSelect(['promotionOfCoreEntity', 'promotionOfImage', 'imageForPromotionOfCoreEntity', 'IDENTITY(promotion.context) AS context_id'])
+            ->addSelect(['relatedLinks', 'promotionOfCoreEntity', 'promotionOfImage', 'imageForPromotionOfCoreEntity', 'IDENTITY(promotion.context) AS context_id'])
+            ->leftJoin('promotion.relatedLinks', 'relatedLinks')
             ->leftJoin('promotion.promotionOfCoreEntity', 'promotionOfCoreEntity')
             ->leftJoin('promotion.promotionOfImage', 'promotionOfImage')
             ->leftJoin('promotionOfCoreEntity.image', 'imageForPromotionOfCoreEntity')
@@ -37,6 +47,7 @@ class PromotionRepository extends EntityRepository
             ->andWhere('promotion.endDate > :datetime')
             ->andWhere('promotion.promotionOfCoreEntity is not null OR promotion.promotionOfImage is not null')
             ->addOrderBy('promotion.weighting', 'ASC')
+            ->addOrderBy('relatedLinks.position', 'ASC')
             ->setFirstResult($offset)
             ->setMaxResults($limit)
             ->setParameter('datetime', $datetime);
--- a/src/Domain/Entity/Promotion.php
+++ b/src/Domain/Entity/Promotion.php
@@ -2,8 +2,10 @@
 
 namespace BBC\ProgrammesPagesService\Domain\Entity;
 
+use BBC\ProgrammesPagesService\Domain\Exception\DataNotFetchedException;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Pid;
 use BBC\ProgrammesPagesService\Domain\ValueObject\Synopses;
+use InvalidArgumentException;
 
 class Promotion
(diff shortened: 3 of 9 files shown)


src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
```diff
@@ -27,8 +27,18 @@ public function findActivePromotionsByContext(array $ancestryIds, DateTimeImmuta
         $contextId = end($ancestryIds);
         $parentIds = array_slice($ancestryIds, 0, -1);
 
+        // YIKES! relatedLinks is a many-to-many join, that could result in an
```
votemike: So promotions won't be shown twice if they have two or more related links?

src/Domain/Entity/Promotion.php
```diff
@@ -28,22 +30,29 @@ class Promotion
...
         PromotableInterface $promotedEntity,
         string $title,
         Synopses $synopses,
         string $url,
         int $weighting,
-        bool $isSuperPromotion
+        bool $isSuperPromotion,
+        ?array $relatedLinks = null
     ) {
+        $this->assertArrayOfType('related links', $relatedLinks, RelatedLink::class);
```
votemike: Would https://secure.php.net/manual/en/function.assert.php be appropriate here?

src/Domain/Entity/Promotion.php
```diff
@@ -28,22 +30,29 @@ class Promotion
     /** @var bool */
     private $isSuperPromotion;
 
+    /** @var RelatedLink[]|null */
+    private $relatedLinks;
```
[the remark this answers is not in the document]
votemike: Ahh, so we can check later.... How come they aren't of type `UnfetchedRelatedLinks`?

src/Mapper/ProgrammesDbToDomain/PromotionMapper.php
```diff
@@ -79,4 +78,19 @@ private function getPromotableModel(array $dbPromotion): PromotableInterface
 
         throw new DataNotFetchedException('All promotions must be joined to CoreEntity and Image');
     }
+
+    private function getRelatedLinksModels(array $dbPromotion, string $key = 'relatedLinks'): ?array
```
votemike: Docblocks for the array of what kind is passed in. And array of what kind is returned?

src/Data/ProgrammesDb/EntityRepository/PromotionRepository.php
BPScott: Correct. Doctrine is smart enough to collapse the two rows that the DB returns into a single promo that looks like `[pid =>'promo', 'relatedLinks' => [['pid=>'rl1'], 'pid => 'rl2']]]`

An example of where we do this elsewhere is `CoreEntityService->findByPidFull('b04v6fz7')`  where you'll get back a single programme with two genres.

src/Data/ProgrammesDb/Entity/Promotion.php
```diff
@@ -119,6 +119,15 @@ class Promotion
     private $cascadesToDescendants = false;
 
     /**
+     * Used in queries only. No getter or setter as creation should be handled
+     * by the RelatedLink entity.
+     *
+     * @ORM\OneToMany(targetEntity="RelatedLink", mappedBy="relatedToPromotion")
+     * @ORM\OrderBy({"position" = "ASC"})
```
BjornTwachtmann: Do we need to specify that here? It could just be a secondary sort on the query, this seems overkill and I don't trust doctrine not to insert random orderby's in places I don't want them.

src/Mapper/ProgrammesDbToDomain/PromotionMapper.php
```diff
@@ -32,14 +32,12 @@ public function getDomainModel(array $dbPromotion): Promotion
...
-            [
-                'promotionOfCoreEntity' => 'CoreEntity',
-                'promotionOfImage' => 'Image',
-            ]
-        );
+        return $this->buildCacheKey($dbPromotion, 'id', [
+            'promotionOfCoreEntity' => 'CoreEntity',
+            'promotionOfImage' => 'Image',
+        ], [
+            'relatedLinks' => 'RelatedLink',
```
BjornTwachtmann: Why the separate array here specifying this as a many/many relationship? Isn't this oneToMany?

src/Domain/Entity/Promotion.php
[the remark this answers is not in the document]
BPScott: When dealing with a property that is an array of things rather than a single entity: null corresponds to "unfetched -  i haven't asked for this data" and `[]` is "I requested this data but there was nothing there".

(contrast dealing with a property that corresponds to a single entity where an Unfetched* entity is "unfetched -  i haven't asked for this data" and null is "I requested this data but there was nothing there")

src/Mapper/ProgrammesDbToDomain/PromotionMapper.php
BPScott: We're returning an array of related links in the `relatedLinks` key, not a single one.

The first array in building the cache keys is for ManyToOne relationships - where the promotionOfImage field relates to a single entity.
The second array in building the cache keys if for OneToMany relationships - where the relatedLinks field is an array.

The logic for building the cache keys our for those two scenarios is slightly different as one is `1` and the other is `[1,2]`

src/Domain/Entity/Promotion.php
BPScott: Neat, I didn't realise that let you throw exceptions in PHP7. Use of assertions sounds like a good plan for these cases where we can't cover our expectations fully using the type system.

We do this in a several other places (e.g. CollapsedBroadcast, CoreEntity, Version). I'd rather make that change to use asserts all in one go rather than piecemeal. Reckon it's worth creating a tech debt ticket to move everything over?

votemike: If you like. Can you pop a link to the ticket in a comment to this thread?

src/Data/ProgrammesDb/Entity/Promotion.php
BPScott: Good point, turns out you can add that to the query rather than being a implicit all-the-time thing.

src/Domain/Entity/Promotion.php
BPScott: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5848

src/Domain/Entity/Promotion.php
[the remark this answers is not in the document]
BPScott: > How come they aren't of type `UnfetchedRelatedLinks`

We can't typehint on "array or UnfetchedRelatedLinks", but we can do `?array`

votemike:
Does Faucet need to be put on int for a few days with these changes before this is merged in?

BPScott:
Faucet doesn't use the Domain or Mapper parts of PagesService - only the Entity models in data/programmesdb/entity. The only thing that deploying to Faucet would exercise would be that the Promotion entity is still valid after we added that property - which is already tested by the unit tests.

PR #168, opened 2017-09-04
CollapsedBroadcastsService

tatitati:
- New test class for method: countUpcomingByProgramme()
- Added additional comments in original service (too much?, is it OK?)
- Added additional tests for some cases (boundaries)

### DIFF
--- a/tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
+++ b/tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
@@ -28,7 +28,7 @@ abstract class AbstractCollapsedBroadcastServiceTest extends AbstractServiceTest
             );
 
             return $this->createConfiguredMock(CollapsedBroadcast::class, [
-                'getServices' => $stubServices
+                'getServices' => $stubServices,
             ]);
         });
         $this->mockServiceRepository = $this->getRepo('ServiceRepository');
--- a/tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
+++ b/tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
@@ -2,6 +2,7 @@
 
 namespace Tests\BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
 
+use BBC\ProgrammesPagesService\Domain\Entity\CollapsedBroadcast;
 use BBC\ProgrammesPagesService\Service\CollapsedBroadcastsService;
 use Tests\BBC\ProgrammesPagesService\AbstractServiceTest;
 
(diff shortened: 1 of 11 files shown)


tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
```diff
@@ -36,4 +43,18 @@ protected function service(): CollapsedBroadcastsService
             $this->mockServiceRepository
         );
     }
+
+    /**
+     * @param Category[] $categories
+     * @return int[]
+     */
+    protected function extractDbAncestryIds(array $categories): array
```
BPScott: This is only used in one test class, is it worth putting it in the abstract test case?

tests/Service/CollapsedBroadcastsService/FilterCategoriesByBroadcastedDateTest.php
```diff
@@ -1,40 +1,71 @@
...
-        $this->assertCount(1, $dbBroadcastedCategories);
-        $this->assertSame([1, 2, 3], $dbBroadcastedCategories[0]->getDbAncestryIds());
+        $this->assertContainsOnly(Genre::class, $categories);
+        $this->assertEquals($expectedCategoriesFoundOnDate, $this->extractDbAncestryIds($categories));
+    }
+
+    public function categoriesBroadcastedOnDateProvider(): array
+    {
+        return [
+            'CASE: none of category specified was not broadcasted on date' => [
```
BPScott: This seems to have the same key as the test case below

Also double negatives sound weird, is there a better way to phrase this?

tests/Service/CollapsedBroadcastsService/FindByCategoryAndStartAtDateRangeTest.php
```diff
@@ -2,36 +2,98 @@
...
-            ->with($ancestry, false, $fromDate, $toDate)
-            ->willReturn($broadcastData);
+            ->willReturn([
+                 ['areWebcasts' => [false, false], 'serviceIds' => [111, 222], 'broadcastIds' => [1, 2, 3, 4]],
+                 ['areWebcasts' => [true, true], 'serviceIds' => [333, 444], 'broadcastIds' => [3, 4, 56, 67]],
+                 ['areWebcasts' => [true, false], 'serviceIds' => [555, 666], 'broadcastIds' => [5, 6, 100]],
+                 ['areWebcasts' => [false, false], 'serviceIds' => [false, false], 'broadcastIds' => [7, 8, 20, 48, 23]],
+                 ['areWebcasts' => [false, false], 'serviceIds' => [true, true], 'broadcastIds' => [8, 9, 12, 122]],
+                 ['areWebcasts' => [false, false], 'serviceIds' => [null, null], 'broadcastIds' => [10, 11]],
+                 ['areWebcasts' => ['0', 0], 'serviceIds' => [777, 888], 'broadcastIds' => [14, 15]],
```
BPScott: Whats with these return values? Can the findByCategoryAncestryAndStartAtDateRange ever return data where areWebcasts is a string or int?

tests/Service/CollapsedBroadcastsService/AbstractCollapsedBroadcastServiceTest.php
tatitati: yeah, you are right

tests/Service/CollapsedBroadcastsService/FilterCategoriesByBroadcastedDateTest.php
tatitati: Yeah!, phpstan was handy here, I just fixed it

BPScott:
Well, approved once you fix the thing phpstan has pointed out

PR #170, opened 2017-09-05
PROGRAMMES-5824 Tests contributionsService

tatitati:


tests/Service/ContributionsService/FindByContributionToProgrammeTest.php
```diff
@@ -2,49 +2,55 @@
...
-
-        $this->mockRepository->expects($this->once())
-            ->method('findByContributionTo')
-            ->with([$dbId], 'programme', false, 5, 10)
-            ->willReturn($dbData);
-
-        $result = $this->service()->findByContributionToProgramme($programme, 5, 3);
-        $this->assertEquals($this->contributionsFromDbData($dbData), $result);
+        return [
+          'CASE: default' => [300, 0, []],
```
votemike: These array keys aren't overly clear to me. Could you expand on them a bit?

tests/Service/ContributionsService/FindByContributionToProgrammeTest.php
```diff
@@ -2,49 +2,55 @@
...
-            ->method('findByContributionTo')
-            ->with([$dbId], 'programme', false, 5, 10)
-            ->willReturn([]);
+        $contributions = $this->service()->findByContributionToProgramme($this->createMock(Programme::class));
 
-        $result = $this->service()->findByContributionToProgramme($programme, 5, 3);
-        $this->assertEquals([], $result);
+        $this->assertCount(count($fakeDbContributions), $contributions);
+        $this->assertContainsOnly(Contribution::class, $contributions);
+        $this->assertEquals($expectedPids, $this->extractPids($contributions));
```
votemike: I'm not overly keen on asserting the return value of a method in this test class equals a whole array. Is there a way to do this that requires less data manipulation?

tests/Service/ContributionsService/FindByContributionToSegmentsTest.php
```diff
@@ -2,61 +2,64 @@
...
+use BBC\ProgrammesPagesService\Domain\Entity\Contribution;
+use BBC\ProgrammesPagesService\Domain\Entity\Segment;
+
 class FindByContributionToSegmentsTest extends AbstractContributionsServiceTest
 {
-    public function testFindByContributionToSegmentsDefaultPagination()
+    /**
+     * @dataProvider paginationProvider
+     */
+    public function testPagination($expectedLimit, $expectedOffset, array $paginationParams)
```
votemike: Typehints

tests/Service/ContributionsService/AbstractContributionsServiceTest.php
```diff
@@ -11,24 +13,26 @@ public function setUp()
...
-        $this->setUpMapper('ContributionMapper', 'contributionFromDbData');
+        $this->setUpMapper('ContributionMapper', function ($dbContribution) {
+            return $this->createConfiguredMock(Contribution::class,
+                ['getPid' => new Pid($dbContribution['pid'])]
+            );
+        });
     }
 
-    protected function contributionsFromDbData(array $entities)
+    protected function service()
```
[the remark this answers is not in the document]
tatitati: Maybe getService().......the service itself is not being mocked. What is being mocked exactly are the dependencies of this service. But that apart is keeping the original funcionallity (which is good! jejeje).

tests/Service/ContributionsService/FindByContributionToProgrammeTest.php
tatitati: I aggree, but not sure if is possible.....we can add multiple assets in rack for each item, can be more transparent, but it also creates more noise with so many asserts....

My idea is to keep it as such. Basically with this we are testing that after creating the entity models (produced really by the mapper but in the tests for us) none of our results are discarded, reordered, etc, so we can continue focusing on the output.
@BPScott any thought?

votemike: Maybe a foreach?

tatitati: Actually that is extractPids:
```php
    protected function extractPids(array $contributions): array
    {
        return array_map(
            function ($contribution) {
                return (string) $contribution->getPid();
            },
            $contributions
        );
    }
```

votemike: But that hides what is happening.

tatitati: maybe we can create something more generic:

```php
function extractProperty(string $property, array $items){
   return ......
}
```

At least this avoid to have extractPids, extractDbIds, ......

votemike: I don't like extracting anything as part of tests

PR #171, opened 2017-09-05
PROGRAMMES-5824 Tests contributorsService

tatitati:
- deleted unnecesary rule to ignore in PHPstan cause is not happening anymore.


votemike:
Does this PR have a ticket? Can you add the ticket reference to the PR title?

PR #174, opened 2017-09-06
PROGRAMMES-5864 Adding interface for Broadcast and BroadcastGap

votemike:
It would be nice to be able to typehint either Broadcasts or Broadcast gaps.  
Do people prefer an interface or an abstract class?


BjornTwachtmann:
I'm not really sure that BroadcastGap even belongs in pages service, it's a frontend concept and is exclusively used there. 
An interface is probably preferable to a base class. I don't think either will actually help that much, as the presenter is still going to want to check whether something's a gap or not when rendering it, but it won't break anything if you want to typehint on it I guess.

felipeparaujo:
I think a interface here fits well

tatitati:
I like the idea

PR #179, opened 2017-09-06
Adding options transformation so we can properly treat dependent options

felipeparaujo:


src/Mapper/ProgrammesDbToDomain/OptionsMapper.php
```diff
@@ -80,4 +85,18 @@ private function getDefaultValues()
 
         return self::$defaults;
     }
+
+    private function transformOptions(array $options): array
+    {
+        // If brand_2016_layout is null (inherit), these options that can only be
```
votemike: Why are we still using this option?

felipeparaujo: Yes, only tv brand pages use the 2016 layout

BPScott: We flicked the flagpole that means now all TV brand pages get the new layout. However eventually we may end up re-using this for radio pages who might want in

PR #182, opened 2017-09-07
PROGRAMMES-5824 tests ProgrammesService

tatitati:


tests/Service/ProgrammesService/CountAllTleosByCategoryTest.php
```diff
@@ -2,21 +2,27 @@
 
 namespace Tests\BBC\ProgrammesPagesService\Service\ProgrammesService;
 
+use BBC\ProgrammesPagesService\Domain\Entity\Genre;
+
 class CountAllTleosByCategoryTest extends AbstractProgrammesServiceTest
 {
-    public function testCountAllTleosByCategory()
+    public function testProtocolWithRepository()
```
BjornTwachtmann: Not quite sure what this test name means, can you make it clearer?

tatitati: Yes of course! jeje

tests/Service/ProgrammesService/CountEpisodeGuideChildrenTest.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $programme = $this->createMock(Programme::class);
+
+        $this->mockRepository
+            ->method('countEpisodeGuideChildren')
+            ->will($this->onConsecutiveCalls(0, 1, 10));
+
+        $this->assertEquals(0, $this->service()->countEpisodeGuideChildren($programme));
+        $this->assertEquals(1, $this->service()->countEpisodeGuideChildren($programme));
+        $this->assertEquals(10, $this->service()->countEpisodeGuideChildren($programme));
```
BjornTwachtmann: I think this style is a little unclear, a dataprovider might be better than having multiple tests in the same method

tatitati: yeah, I was expecting a possible comment about it. OK, I will, I aggree.

BjornTwachtmann:
Other than those two comments, looks good

PR #183, opened 2017-09-07
Tests versionsService

tatitati:


phpstan.neon
```diff
@@ -14,7 +14,6 @@ parameters:
         - '#Parameter \#6 \$endAt of class BBC\\ProgrammesPagesService\\Domain\\Entity\\Broadcast constructor expects DateTimeImmutable, DateTimeImmutable\|null given.#'
         # PHPUnit ignores
         - '#PHPUnit_Framework_MockObject_MockObject given.#'
-        - '#but returns PHPUnit_Framework_MockObject_MockObject.#'
```
tatitati: from now this rule doesn't need to be ignored is saying PHPstan because his case is not found anymore

PR #185, opened 2017-09-08
tests cleaned dead code

tatitati:


tests/AbstractServiceTest.php
```diff
@@ -8,9 +8,7 @@
 
 abstract class AbstractServiceTest extends TestCase
 {
-    const ENTITY_NS = 'BBC\ProgrammesPagesService\Domain\Entity\\';
     const REPOSITORY_NS = 'BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\\';
```
felipeparaujo: you got rid of the other namespaces, can't you do the same for this one?

tatitati: Yes!, is my next PR (is ready)

felipeparaujo: ah great news. this is good to go then.

PR #189, opened 2017-09-13
Adding image service

votemike:
Required for https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5734


src/Data/ProgrammesDb/EntityRepository/ImageRepository.php
```diff
@@ -0,0 +1,19 @@
...
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Query;
+
+class ImageRepository extends EntityRepository
+{
+    public function findByPid(string $pid): ?array
+    {
+        $qb = parent::createQueryBuilder('image')
+            ->andWhere('image.pid = :pid')
+            ->andWhere('image.isEmbargoed = 0')
```
BPScott: This should be automatically added to all embargoable entities by the [EmbargoedFilter](https://github.com/bbc/programmes-pages-service/blob/master/src/Data/ProgrammesDb/Filter/EmbargoedFilter.php), so you don't need to explicitly specify it.

src/Service/ImagesService.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public function __construct(
+        ImageRepository $repository,
+        ImageMapper $mapper,
+        CacheInterface $cache
+    ) {
+        parent::__construct($repository, $mapper, $cache);
+    }
+
+    public function findByPid(Pid $pid, $ttl = CacheInterface::NORMAL): ?Image
```
BPScott: Did you say there might be multiple iSite promos? If so would we want to do findByPids so that we can request multiple images all in one query?

src/Data/ProgrammesDb/EntityRepository/ImageRepository.php
votemike: Is that automatic? Or do I need to add a trait or annotation or something?

src/Service/ImagesService.php
votemike: There can be more than one. But I haven't found an instance when we want to query for more than one. I think we should wait until we need that before we add it.

src/Data/ProgrammesDb/EntityRepository/ImageRepository.php
```diff
@@ -0,0 +1,18 @@
...
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository;
+
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Query;
+
+class ImageRepository extends EntityRepository
+{
+    public function findByPid(string $pid): ?array
+    {
+        $qb = parent::createQueryBuilder('image')
```
votemike: Is this the simplest that this method can get? I feel this is so generic that it can be refactored even more.

src/Data/ProgrammesDb/EntityRepository/ImageRepository.php
BPScott: It's automatic based on if an entity has a field called isEmbargoed yep

src/Data/ProgrammesDb/EntityRepository/ImageRepository.php
BPScott: It is very rare that we need just the entity without any joins - most of the time we want a FindByPidFull that joins to a set of fields that differs on a per entity basis (e.g. CoreEntity joins to Image & Masterbrand, Contribution join to Contributor etc).

I don't think it's worth creating an abstract thing for this, as 90% of the time it is too constrained to be useful.

PR #192, opened 2017-09-14
Networks with all services

felipeparaujo:


src/Domain/Entity/Network.php
```diff
@@ -63,7 +66,8 @@ public function __construct(
         bool $isChildrens = false,
         bool $isWorldServiceInternational = false,
         bool $isInternational = false,
-        bool $isAllowedAdverts = false
+        bool $isAllowedAdverts = false,
+        ?array $services = null
```
votemike: Do we want to validate this as it goes in?

src/Mapper/ProgrammesDbToDomain/NetworkMapper.php
```diff
@@ -18,6 +18,8 @@ public function getCacheKey(array $dbNetwork): string
         return $this->buildCacheKey($dbNetwork, 'id', [
             'image' => 'Image',
             'defaultService' => 'Service',
+        ], [
+            'services' => 'Service',
```
votemike: Services instead of Service?

votemike:
And some tests?

src/Mapper/ProgrammesDbToDomain/NetworkMapper.php
BjornTwachtmann: This is correct, anything else will not get the right cache key

src/Domain/Entity/Network.php
```diff
@@ -73,6 +78,18 @@ public function __construct(
             ));
         }
 
+        // Validate array of Services
+        if (is_null($services)) {
```
BjornTwachtmann: I don't think that's going to work :)

felipeparaujo: no, it broke everything. but i unbroke it, so it's all good now

tests/Data/ProgrammesDb/EntityRepository/ServiceRepository/FindByIdsWithNetworkServicesListTest.php
```diff
@@ -0,0 +1,54 @@
...
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ServiceRepository;
+use BBC\ProgrammesPagesService\Domain\ValueObject\Sid;
+use Tests\BBC\ProgrammesPagesService\AbstractDatabaseTest;
+
+/**
+ * @covers BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ServiceRepository::<public>
+ */
+class FindByIdsWithNetworkServicesListTest extends AbstractDatabaseTest
+{
+    /** @var  ServiceRepository */
```
votemike: TOO MANY SPACES!!!!!

felipeparaujo: jesus, i'm sorry, i'm sorry!

votemike: Don't be sorry to Jesus. Be sorry to me.

src/Data/ProgrammesDb/Entity/Network.php
```diff
@@ -135,6 +135,13 @@ class Network
      */
     private $endDate;
 
+    /**
+     * @var array
```
BPScott: This isn't an array, it's a Doctrine Collection or something like that.

src/Domain/Entity/Network.php
```diff
@@ -186,4 +204,18 @@ public function getOption(string $key)
...
     }
+
+    /**
+     * @throws DataNotFetchedException
+     */
+    public function getServices(): array
+    {
+        if (is_null($this->services)) {
+            throw new DataNotFetchedException(
+                'Could not get Services of Network "' . $this->nid . '" as it was not fetched'
```
BPScott: "..as they were not fetched" would be more gramatical

src/Mapper/ProgrammesDbToDomain/NetworkMapper.php
BPScott: The key of this array corresponds to the property name in the Data Entity. The value is the name of the Mapper that knows how to build this item - in this case the `ServiceMapper`.

tests/Data/ProgrammesDb/EntityRepository/ServiceRepository/FindByIdsWithNetworkServicesListTest.php
```diff
@@ -0,0 +1,54 @@
...
+    public function setUp()
+    {
+        $this->loadFixtures(['NetworksFixture']);
+        $this->repo = $this->getEntityManager()->getRepository('ProgrammesPagesService:Service');
+    }
+
+    public function testFindServiceBy()
+    {
+        $sid = new Sid('bbc_radio_fourlw');
+        $dbId = $this->getDbId($sid);
```
BPScott: Use the utility functions rather than wrangling getSQLLogger bits yourself:

Also you don't need the Sid value object at all as this is a DB query so it's all strings.

```
$dbId = $this->getDbIdFromPersistentIdentifier('bbc_radio_fourlw', 'Service', 'sid');
```

Or actually your Service fixture has a PID so use that instead of the Sid (as that's the id that `getDbIdFromPersistentIdentifier` defaults to using):

```
$dbId = $this->getDbIdFromPersistentIdentifier('p00fzl7k', 'Service');
```

tatitati:
is ready to merge this PR?, I need it for another ticket

PR #195, opened 2017-10-04
PROGRAMMES-5802 aggregated_galleries_count

Tasiobg:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5802
Depends on https://github.com/bbc/programmes-faucet/pull/29

- update and rename variables and functions for "available galleries count" to "aggregated galleries count"
- Add aggregatedGalleriesCount to Clips
- unitest


src/Data/ProgrammesDb/Entity/Clip.php
```diff
@@ -2,11 +2,13 @@
...
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\GalleriesCountableInterface;
 use Doctrine\ORM\Mapping as ORM;
 
 /**
  * @ORM\Entity()
  */
-class Clip extends ProgrammeItem
+class Clip extends ProgrammeItem implements GalleriesCountableInterface
 {
+    use Traits\AggregatedGalleriesCountMethodsTrait;
```
votemike: Can this be imported?

src/Domain/Entity/Episode.php
```diff
@@ -18,7 +18,7 @@ class Episode extends ProgrammeItem
     private $availableClipsCount;
 
     /** @var int */
-    private $availableGalleriesCount;
+    private $aggregatedGalleriesCount;
```
votemike: Would it not be better to put all of these changes in `ProgrammeItem`?

src/Domain/Entity/ProgrammeContainer.php
```diff
@@ -21,7 +21,7 @@
     private $availableEpisodesCount;
 
     /** @var int */
-    private $availableGalleriesCount;
+    private $aggregatedGalleriesCount;
```
votemike: Would it not be better to put all of these changes in `Programme`?

src/Data/ProgrammesDb/Entity/Clip.php
Tasiobg: We are using traits like that in few places, I think importing them is a more elegant, thanks.

src/Domain/Entity/ProgrammeContainer.php
Tasiobg: Looks like the aggregatedGalleriesCount code can be moved inside Programme, thanks.

PR #198, opened 2017-10-06
Make fewer ancestry queries through the magic of caching

BjornTwachtmann:
STOP MAKING THE SAME SQL QUERY 4 TIMES ON THE SAME PAGE. BUT NOT SHOUTING.

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -561,15 +561,7 @@ QUERY;
         $this->ancestryCache = [];
     }
 
-    private function resolveParents(array $programmes): array
-    {
-        return $this->abstractResolveAncestry(
-            $programmes,
-            [$this, 'programmeAncestryGetter']
-        );
-    }
-
-    private function programmeAncestryGetter(array $ids): array
+    public function coreEntityAncestryGetter(array $ids): array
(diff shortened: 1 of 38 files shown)


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -186,13 +188,17 @@ public function findByPidFull(string $pid, string $entityType = 'CoreEntity'): ?
             ->setParameter('pid', $pid);
 
         $result = $qb->getQuery()->getOneOrNullResult(Query::HYDRATE_ARRAY);
-        $withHydratedParents = $result ? $this->resolveParents([$result])[0] : $result;
-        return $withHydratedParents ? $this->resolveCategories([$withHydratedParents])[0] : $withHydratedParents;
+        if (!$result) {
+            return $result;
+        }
+        $this->addToAncestryCache([$result]);
```
BPScott: I'm guessing this is in findByPid full, because resolveParents does not store the current base programme in the cache (only its ancestry), and on things like the tlec page we are getting lots of things that are immediate children of this top level item, thus we can sidestep future ancestry requests for them by having this parent already stored.

BjornTwachtmann: You are technically correct. The best kind of correct.

PR #202, opened 2017-10-10
Fixing bug when episode would not be retrieved in the streamable from…

votemike:
PROBABLY INSTEAD OF UPDATING THIS METHOD, DEPRECATE IT. TBC.

I think this is needed as a part of fixing https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5737?focusedCommentId=3605286&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3605286

Relates to this PR https://github.com/bbc/programmes-frontend/pull/234


BjornTwachtmann:
This seems wrong to me. Maybe I'm not understanding something, but filtering by streamable_until >= NOW() will only get you the set of streamable episodes that are not streamable indefinitely (i.e. they have an expiry date). 

Why do you need that set for the linked ticket?

votemike:
@BjornTwachtmann the aim is to get the first item that has not expired yet but is also not streamable.

src/Service/ProgrammesAggregationService.php
```diff
@@ -54,6 +54,7 @@ public function findStreamableDescendantEpisodes(
 
     /**
      * @return Episode[]
+     * @deprecated Not actually needed
```
BPScott: Should we fire trigger_error here too so then people know what to change?

BjornTwachtmann:
Don't mark these as deprecated. They're not used in clifton/faucet so just remove them here and in frontend, if it's dead code

PR #204, opened 2017-10-16
DO NOT REVIEW Adding upcoming streamable episode count

votemike:
Needed for https://github.com/bbc/programmes-frontend/pull/243

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -33,6 +33,34 @@ class CoreEntityRepository extends MaterializedPathRepository
 
     private $ancestryCache = [];
 
+    /**
+     * @param int[] $ancestryDbIds
+     * @param string $entityType
+     * @param DateTimeImmutable $fromDateTime
+     * @return int
+     */
+    public function countUpcomingStreamableDescendantsByType(array $ancestryDbIds, string $entityType, DateTimeImmutable $fromDateTime): int
+    {
+        $this->assertEntityType($entityType, ['Clip', 'Episode']);
+
+        $qb = $this->createQueryBuilder('entity')
+            ->select('COUNT(entity)')
+            ->from('ProgrammesPagesService:' . $entityType, 'entity')
+            ->leftJoin('entity.masterBrand', 'masterBrand')
+            ->leftJoin('masterBrand.network', 'network')
+            ->leftJoin('entity.image', 'image')
+            ->leftJoin('masterBrand.image', 'mbImage')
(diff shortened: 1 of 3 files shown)


src/Service/ProgrammesAggregationService.php
```diff
@@ -30,6 +30,11 @@ public function __construct(
         parent::__construct($repository, $mapper, $cache);
     }
 
+    public function countUpcomingStreamableDescendantEpisodes(Programme $programme): int
```
votemike: I don't actually need the count in frontend, I want to see if any exist.
Should I also allow a limit to be passed in so that the query stops once it has found the first instance?
Or possible have a bool called `$exists` that limits the query to 1?

Tasiobg: In terms of perfomance should be fast to use a LIMIT in a query but you can't use it inside a COUNT, should be a normal SELECT, in this case you may want to rename the function and return a boolean.

votemike: Why can't you use LIMIT in a COUNT?

Tasiobg: @votemike this is a good explanation https://stackoverflow.com/questions/17020842/mysql-count-with-limit

votemike: I always thought you could use LIMIT with a COUNT.... Oops.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -33,6 +33,34 @@ class CoreEntityRepository extends MaterializedPathRepository
...
+     * @param int[] $ancestryDbIds
+     * @param string $entityType
+     * @param DateTimeImmutable $fromDateTime
+     * @return int
+     */
+    public function countUpcomingStreamableDescendantsByType(array $ancestryDbIds, string $entityType, DateTimeImmutable $fromDateTime): int
+    {
+        $this->assertEntityType($entityType, ['Clip', 'Episode']);
+
+        $qb = $this->createQueryBuilder('entity')
```
felipeparaujo: You shouldn't be using `$this->createQueryBuilder`, as you're changing the `select()` and `from()` bits. Something like `$this->getEntityManager()->createQueryBuilder()` would be more suitable.

votemike: That's what it is.... isn't it?

src/Service/ProgrammesAggregationService.php
votemike: So 3 options:
1.  SELECT with a LIMIT
2. Keep it as a COUNT
3. Use EXISTS

PR #206, opened 2017-10-17
Changing streamable episode ordering

votemike:
To fix this problem: https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5737?focusedCommentId=3613303&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3613303

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -232,18 +232,20 @@ QUERY;
     }
 
     /**
-     * @param int[]    $ancestryDbIds
-     * @param string   $entityType
+     * @param int[] $ancestryDbIds
+     * @param string $entityType
+     * @param DateTimeImmutable $now
      * @param int|null $limit
-     * @param int      $offset
+     * @param int $offset
      * @return array
      */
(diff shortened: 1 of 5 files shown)


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -232,28 +232,32 @@ public function findDescendantsByType(array $ancestryDbIds, string $entityType,
...
+     * @param int[] $ancestryDbIds
+     * @param string $entityType
+     * @param DateTimeImmutable $now
      * @param int|null $limit
-     * @param int      $offset
+     * @param int $offset
      * @return array
      */
-    public function findStreamableDescendantsByType(array $ancestryDbIds, string $entityType, ?int $limit, int $offset) : array
+    public function findStreamableDescendantsByType(array $ancestryDbIds, string $entityType, DateTimeImmutable $now, ?int $limit, int $offset) : array
```
felipeparaujo: I would call this cutoffTime to be in line with the rest of the stuff

votemike: But it's not the cutoff time. I could do the 7 day logic in the Service I suppose. Then pass it through?

votemike:
@BPScott not keen on those sort of tests. Would that not be better as an automated test by the testers?

votemike:
Test this please

PR #210, opened 2017-10-20
Added new field 'contact_details' in core_entity/options column 

tatitati:
[PROGRAMMES-5933](https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5933)
[Faucet PR](https://github.com/bbc/programmes-faucet/pull/33)


src/Domain/Entity/ContactMediaMap.php
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @return string[]     Example: ['reddit', 'email']
+     */
+    public function getNamesOfUsedMedia(): array
+    {
+        return array_keys($this->contactsMapByMedia);
+    }
+
+    /**
+     * @return ContactMedia[] | []
```
votemike: `@return ContactMedia[]`

src/Domain/Entity/ContactMediaMap.php
```diff
@@ -0,0 +1,52 @@
...
+class ContactMediaMap
+{
+    /**
+     * Example:
+     *  [
+     *      'email'    => [<ContactMedia>, <ContactMedia>, <ContactMedia>],
+     *      'facebook' => [<ContactMedia>],
+     *  ]
+     *
+     * @var mixed[]
```
votemike: `@var ContactMedia[][]`?

src/Domain/ValueObject/ContactMedia.php
```diff
@@ -0,0 +1,41 @@
...
+
+use BBC\ProgrammesPagesService\Domain\Enumeration\ContactMediaEnum;
+
+class ContactMedia
+{
+    private $value;
+    private $freetext;
+    private $type;
+
+    public function __construct($type, $value, $freetext)
```
votemike: Typehinting?
Should you check the type is valid?

src/Domain/Entity/ContactMediaMap.php
tatitati: true

src/Domain/Entity/ContactMediaMap.php
tatitati: I prefer to put an example. That kind of [][]....still is so blur...that for instance doesn't say anything about if we use numerical/string keys.

src/Domain/ValueObject/ContactMedia.php
tatitati: Yeah, good point. Currently and checking that in the src/Domain/Entity/ContactMediaMap.php (which is orchestring the creation of ContactMedia), but I agree with you, it should be here.

src/Domain/Entity/ContactMediaMap.php
votemike: I'm not saying to get rid of the example. Keep it if you like. Just change `@var mixed[]`

src/Mapper/ProgrammesDbToDomain/OptionsMapper.php
```diff
@@ -62,12 +65,16 @@ public function getDomainModel(array $options, array ...$parentEntities)
 
         // set default values
         $defaults = $this->getDefaultValues();
-        foreach ($defaults as $key => $value) {
+        foreach ($defaults as $key => $defaultValue) {
```
votemike: Can this not be done with array_merge or something like that?

src/Domain/Entity/ContactMediaMap.php
```diff
@@ -0,0 +1,58 @@
...
+        $validsMedia = ContactMediaEnum::VALID_MEDIA;
+
+        if (!isset($validsMedia[$contactMediaType])) {
+            throw new InvalidArgumentException('Trying to get a contact detail with unknown type');
+        }
+
+        return $this->contactsMapByMedia[$contactMediaType] ?? [];
+    }
+
+    public function addContactMedia(ContactMedia $contactMedia): self
```
BPScott: Domain objects should be immutable. This means we don't have to worry about if they have been modified at any point in their lifecycle. I believe this object is the first that breaks that rule

Can you remove this `addContactMedia`, and instead construct the list of valid items within the Mapper.

src/Domain/Entity/ContactMediaMap.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function getNamesOfUsedMedia(): array
+    {
+        return array_keys($this->contactsMapByMedia);
+    }
+
+    /**
+     * @return ContactMedia[]
+     */
+    public function getContactsByMedia(string $contactMediaType): array
```
BPScott: When will we need to get all contact info grouped by the type?

In v2 I don't think we display all `email`s together for example, we just display things in the order that they were defined in iSite. Is there some corner case I'm not aware of?

src/Domain/Enumeration/ContactMediaEnum.php
```diff
@@ -0,0 +1,44 @@
...
+    public const PINTEREST = 'pinterest';
+    public const TUMBLR = 'tumblr';
+    public const STUMBLE_UPON = 'stumble_upon';
+    public const LINKEDIN = 'linkedin';
+    public const REDDIT = 'reddit';
+    public const DIGG = 'digg';
+    public const INSTAGRAM = 'instagram';
+    public const OTHER = 'other';
+
+    public const VALID_MEDIA = [
```
BPScott: This way of checking valid values differs from how we do it our other enums. Checking a key is faster than checking through an array (O(1) vs O(n)) but we shouldn't try and avoid having multiple ways of doing a single thing.

Having things that consume enums check if a value is valid in differing ways differ based on the internal data structure (`isset(Enum::VALID_MEDIA[$value])` here vs `in_array($value, Enum::validValues())` in other enums) feels a bit janky.

Perhaps the long term solution is to have Enums expose a static isValid method:

```php

private const VALID_VALUES = [/*...*/];

public static isValid($value): bool {
  return isset(self::VALID_VALUES);
}
```


We can then update the other classes to have this isValid convention in a separate PR. What do you think?

BPScott: Even more thoughts: Why do we even need to validate at this phase anyway? Surely any validation would have occurred on ingest in Faucet and we know we've got good information within the database.

src/Domain/Entity/ContactMediaMap.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $validsMedia = ContactMediaEnum::VALID_MEDIA;
+
+        if (!isset($validsMedia[$contactMediaType])) {
+            throw new InvalidArgumentException('Trying to get a contact detail with unknown type');
```
BPScott: I don't think we should throw an error here and instead power through, not adding an invalid thing to the list.

We don't want a page to start throwing 500 errors because some invalid thing snuck its way into isite.

PR #211, opened 2017-10-20
Add findByPids

mattgreenham:
Added findByPids to ProgrammesService, for RecEng, which returns a list of pids. This enables us to get an array of Programmes for an array of pids

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -199,9 +199,9 @@ QUERY;
     /**
      * @param string[] $pids
      * @param string $entityType
-     * @return array|null
+     * @return array
      */
-    public function findByPids(array $pids, string $entityType = 'CoreEntity'): ?array
+    public function findByPids(array $pids, string $entityType = 'CoreEntity'): array
     {
         $this->assertEntityType($entityType, self::ALL_VALID_ENTITY_TYPES);
         $qb = $this->getEntityManager()->createQueryBuilder()
@@ -216,11 +216,8 @@ QUERY;
             ->andWhere('entity.pid IN (:pids)')
             ->setParameter('pids', $pids);
 
-        $result = $qb->getQuery()->getResult();
-        if (!$result) {
-            return $result;
-        }
-        return $this->resolveParents([$result]);
+        $result = $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
+        return $this->resolveParents($result);
     }
 
     public function findByIds(array $ids): array
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -196,6 +196,33 @@ QUERY;
         return $this->resolveCategories($withHydratedParents)[0];
     }
 
+    /**
+     * @param string[] $pids
+     * @param string $entityType
+     * @return array|null
+     */
+    public function findByPids(array $pids, string $entityType = 'CoreEntity'): ?array
+    {
+        $this->assertEntityType($entityType, self::ALL_VALID_ENTITY_TYPES);
+        $qb = $this->getEntityManager()->createQueryBuilder()
+            ->select(['entity', 'image', 'masterBrand', 'network', 'mbImage', 'category', 'nwImage'])
+            ->from('ProgrammesPagesService:' . $entityType, 'entity') // For filtering on type
+            ->leftJoin('entity.image', 'image')
+            ->leftJoin('entity.masterBrand', 'masterBrand')
+            ->leftJoin('masterBrand.image', 'mbImage')
+            ->leftJoin('masterBrand.network', 'network')
+            ->leftJoin('network.image', 'nwImage')
+            ->leftJoin('entity.categories', 'category')
+            ->andWhere('entity.pid IN (:pids)')
+            ->setParameter('pids', $pids);
+
+        $result = $qb->getQuery()->getResult();
+        if (!$result) {
+            return $result;
+        }
+        return $this->resolveParents([$result]);
+    }
+
     public function findByIds(array $ids): array
     {
         $results = $this->createQueryBuilder('programme')
--- a/src/Service/ProgrammesService.php
+++ b/src/Service/ProgrammesService.php
@@ -207,15 +207,23 @@ class ProgrammesService extends AbstractService
     {
         $this->assertEntityType($entityType, self::ALL_VALID_ENTITY_TYPES);
 
-        $pids = array_map('strval', $pids); // convert to strings
+        $validPids = [];
 
-        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode('/', $pids), $entityType, $ttl);
+        foreach ($pids as $pid) {
+            if ($pid instanceof Pid) {
+                $validPids[] = (string) $pid;
+            } else {
+                throw new InvalidArgumentException("Called findByPids with an invalid type. Array must contain only Pids.");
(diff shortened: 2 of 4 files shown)


tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindByPidsTest.php
```diff
@@ -0,0 +1,69 @@
...
+    {
+        $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity')->clearAncestryCache();
+    }
+
+    public function testFindByPids()
+    {
+        $this->loadFixtures(['MongrelsFixture']);
+        $repo = $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity');
+
+        $entity = $repo->findByPids(['b00swyx1']);
```
BjornTwachtmann: Probably worth adding multpiple PIDs to this array to test that, we've already got findByPid if we want a single one

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -196,6 +196,30 @@ public function findByPidFull(string $pid, string $entityType = 'CoreEntity'): ?
...
+        $qb = $this->getEntityManager()->createQueryBuilder()
+            ->select(['entity', 'image', 'masterBrand', 'network', 'mbImage', 'category', 'nwImage'])
+            ->from('ProgrammesPagesService:' . $entityType, 'entity') // For filtering on type
+            ->leftJoin('entity.image', 'image')
+            ->leftJoin('entity.masterBrand', 'masterBrand')
+            ->leftJoin('masterBrand.image', 'mbImage')
+            ->leftJoin('masterBrand.network', 'network')
+            ->leftJoin('network.image', 'nwImage')
+            ->leftJoin('entity.categories', 'category')
+            ->andWhere('entity.pid IN (:pids)')
```
votemike: Do we want to limit the number of pids that can be passed in? I think the most you can have is 65535.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -196,6 +196,30 @@ public function findByPidFull(string $pid, string $entityType = 'CoreEntity'): ?
...
+            ->setParameter('pids', $pids);
```
votemike: What if the array contains something other that strings?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -196,6 +196,30 @@ public function findByPidFull(string $pid, string $entityType = 'CoreEntity'): ?
         return $this->resolveCategories($withHydratedParents)[0];
     }
 
+    /**
+     * @param string[] $pids
+     * @param string $entityType
+     * @return array
+     */
+    public function findByPids(array $pids, string $entityType = 'CoreEntity'): array
```
votemike: What if you pass in 4 pids, but only 3 are found. Do we want to error or anything?

src/Service/ProgrammesService.php
```diff
@@ -197,6 +197,30 @@ function () use ($pid, $entityType) {
...
+     * @param Pid[] $pids
+     * @param string $entityType
+     * @param string $ttl
+     * @return Programme[]
+     */
+    public function findByPids(array $pids, string $entityType = 'Programme', $ttl = CacheInterface::NORMAL)
+    {
+        $this->assertEntityType($entityType, self::ALL_VALID_ENTITY_TYPES);
+
+        $pids = array_map('strval', $pids); // convert to strings
```
votemike: You're not making sure the $pids are all `Pids`.

src/Service/ProgrammesService.php
```diff
@@ -197,6 +197,30 @@ function () use ($pid, $entityType) {
...
+
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, implode('/', $pids), $entityType, $ttl);
```
votemike: If you check `findGenreByUrlKeyAncestry()` it seems we implode by pipe (`|`)

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
votemike: Seeing as it is so similar, is it worth refactoring this and `findByPid` to use as much of the same code as possible. Instead of duplicating?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: I don't think it needs to be explicitly limited. If somebody's passing that many PIDs here, that's already going to make the page unusably slow, anything like that should be caught in code review.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: No, I think that's legitmate behaviour that the calling function should handle.

BjornTwachtmann: TBH I deliberately avoided that. findByPid is about the most called function in the codebase, and it actually has some subtly different behaviour that I don't want people breaking if they touch this function. I'll eat the overhead of the duplication to avoid people breaking findByPid

mattgreenham: I think it's fine if we're consistent with the behaviour of findByPid and just return what is found & return an empty array if nothing.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: This function is for internal use in pages service, the input should be checked in the Service call, I'm reluctant to duplicate that

src/Service/ProgrammesService.php
BjornTwachtmann: Yeah, this should probably be asserted here, you're right

src/Service/ProgrammesService.php
```diff
@@ -197,6 +197,38 @@ function () use ($pid, $entityType) {
...
+     * @return Programme[]
+     */
+    public function findByPids(array $pids, string $entityType = 'Programme', $ttl = CacheInterface::NORMAL)
+    {
+        $this->assertEntityType($entityType, self::ALL_VALID_ENTITY_TYPES);
+
+        $validPids = [];
+
+        foreach ($pids as $pid) {
+            if ($pid instanceof Pid) {
```
votemike: Do we assert an array only contains instances of something any where else in PPS? I feel this should be genericied.

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BPScott: An example of this occuring is if a programme is embargoed. In this case we'd want to return the programmes without the embargoed item rather than error. Adding an embargoed item somewhere may be an editorial decision so it appears as soon as the embargo is lifted. No need to punish the user for that editorial workflow.

src/Service/ProgrammesService.php
BPScott: A few places yeah. I believe you raised a ticket to look into an Assert library that offers these generic checks

votemike: Did I? Well done past Mike.

BPScott: Ah no, I lied, I think you mentioned it to me then I raised https://jira.dev.bbc.co.uk/browse/PROGRAMMES-5848

PR #212, opened 2017-10-23
updated ref isite options entity

tatitati:
We want to ingest now contact data from isite. So far we only have ```admin``` and ```local```, this PR is to support a new type: ```contact```

### DIFF
--- a/src/Data/ProgrammesDb/Entity/RefIsiteOptions.php
+++ b/src/Data/ProgrammesDb/Entity/RefIsiteOptions.php
@@ -15,6 +15,7 @@ class RefIsiteOptions
 {
     const TYPE_LOCAL = 'local';
     const TYPE_ADMIN = 'admin';
+    const TYPE_CONTACT = 'contact';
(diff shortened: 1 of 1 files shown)


src/Data/ProgrammesDb/Entity/RefIsiteOptions.php
```diff
@@ -63,11 +64,11 @@ class RefIsiteOptions
     private $fileId;
 
     /**
-     * local|admin
+     * local|admin|contact
      *
      * @var string
      *
-     * @ORM\Column(type="string", length=5, nullable=false)
+     * @ORM\Column(type="string", length=7, nullable=false)
```
felipeparaujo: leave this as 35, 7 is only fixing the problem for now. There's no problem having this as a longer string

tatitati: yeah, I thought that as well. Cool, I will.

PR #215, opened 2017-11-15
PROGRAMMES-6023 - Caching Null results for some queries

felipeparaujo:


src/Service/CollapsedBroadcastsService.php
```diff
@@ -137,7 +137,8 @@ function () use ($programme) {
      */
     public function findNextDebutOrRepeatOnByProgrammeWithFullServicesOfNetworksList(
         Programme $programme,
-        $ttl = CacheInterface::NORMAL
+        $ttl = CacheInterface::NORMAL,
+        $nullTtl = CacheInterface::NORMAL
```
BjornTwachtmann: 60 seconds seems safer here

PR #216, opened 2017-11-17
Schema change for On Demand Sort Date. Changes to PartialDate.

BjornTwachtmann:
Don't merge yet.


src/Domain/ValueObject/PartialDate.php
```diff
@@ -22,9 +23,10 @@ class PartialDate implements JsonSerializable
      */
     public function __construct(int $year, int $month = 0, int $day = 0)
     {
+        list($checkMonth, $checkDay) = $this->normaliseMonthAndDay($month, $day);
```
felipeparaujo: pretty petty, don't change it if you don't want to, but I would do `[$checkMont, $checkDay]`

PR #217, opened 2017-11-21
PROGRAMMES-6052 findOnNowByService cache issues

Tasiobg:
- Set max cache time to 300s
- Add function flushOnNowByService so we can flush it when an stale object still in memory and is not longer valid

Frontend changes to use flushOnNowByService
https://github.com/bbc/programmes-frontend/commit/01d5685fbec0814446ce3797f1527d13919d9d0c


tatitati:
Looks ok 👍

PR #218, opened 2017-11-29
PROGRAMMES-6061

Tasiobg:
- get network if there is not defaultService when calling findByUrlKeyWithDefaultService

Requires https://github.com/bbc/programmes-frontend/pull/310


votemike:
Why has this been done?

Tasiobg:
@votemike Because we found that networks like "news" don't have default service and this was trowing 404 on the "on-now", example: schedules/network/bbcone/on-now?partial=legacy_2013
Now we handle this better here: https://github.com/bbc/programmes-frontend/pull/310/files

PR #225, opened 2018-01-25
Update collapsed broadcast service to save queries for service lists

BjornTwachtmann:


votemike:
Do we ever want the services back in the order they went in?

votemike:
This will only save a query if ALL the serviceIds requested match ALL of the ones previously requested. Right?

src/Service/CollapsedBroadcastsService.php
```diff
@@ -783,16 +785,57 @@ function ($memo, $broadcast) {
...
+    }
 
-        // Fetch all the used services, keyed by their id
+    private function joinServiceIds(array $serviceIds)
+    {
+        sort($serviceIds, SORT_NUMERIC);
+        return join(",", $serviceIds);
+    }
+
+    private function keyServiceArrayById(array $services)
```
votemike: Does this return an array of services?

BjornTwachtmann:
Very belated response to the above: Yes, this only saves a query if all the services are the same. However that's an extremely common case (all the services will be the same for anything that's only on BBC one, for example).

PR #227, opened 2018-01-30
Order CoreEntityRepository->findByPids results based on pids passed in

BPScott:
SQL makes no claims about the order of what it returns. Make sure
findByPids(['pid1','pid3','pid2']) returns the entities represented by
those pids in that order.

I'm fairly sure it makes more sense to do this ordering here rather than in ProgrammesService->findByPids. What do you think?

### DIFF
--- a/tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindByPidsTest.php
+++ b/tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindByPidsTest.php
@@ -19,14 +19,22 @@ class FindByPidsTest extends AbstractDatabaseTest
         $this->loadFixtures(['MongrelsFixture']);
         $repo = $this->getEntityManager()->getRepository('ProgrammesPagesService:CoreEntity');
 
-        $entity = $repo->findByPids(['b00swyx1', 'b010t150']);
-        $this->assertInternalType('array', $entity);
+        $pids = ['b00tf1zy', 'b010t150', 'b00swyx1'];
 
-        $this->assertSame('b00swyx1', $entity[0]['pid']);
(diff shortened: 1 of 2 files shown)


tests/Data/ProgrammesDb/EntityRepository/CoreEntityRepository/FindByPidsTest.php
```diff
@@ -19,14 +19,22 @@ public function testFindByPids()
...
         // findByPids query and parent lookup query only
         $this->assertCount(2, $this->getDbQueries());
+
+        // A similar query but with a different order. This asserts that the
+        // entities that are returned are in the same order as the $pids array
+        $pids = ['b010t150', 'b00swyx1', 'b00tf1zy'];
+
+        $entity = $repo->findByPids($pids);
+        $this->assertInternalType('array', $entity);
+        $this->assertSame($pids, array_column($entity, 'pid'));
```
votemike: And another test with an embardgoed entity or something?

votemike:
When do we need entities in this order? For recommendations?

BjornTwachtmann:
Seems simple enough. Not sure what it's needed for though?

BPScott:
Consider the recommendations API, which returns a list of PIDs ordered based on relevance. We then pass that list of PIDs into findByPids() to turn them into Programme objects, however prior to this PR findByPids() does not maintain the order, which results in entities swapping positions - no longer maintaining that "ordered by relevance".

I think here is the best place to solve this by making findByPids result ordering consistent with the order of the pids passed into it, rather than having to check that in every place that need to call it (which currently is only in the recommendations ExternalApi but I can see this assumption biting us in the future

PR #228, opened 2018-02-20
Allow RelatedLinksService->findByRelatedToProgramme() filter by link type

BPScott:
Allow for passing in an array of related link types.

I don't think filter by type requires an index, as we're already
filtering by the related_to id, which reduces the result set down to
less than 50 results in all cases (and less than 20 results in all but
45 cases).

Fixes PROGRAMMES-6221.

---

This adds a parameter in the middle of the arg list for `findByRelatedToProgramme`, which is technically a breaking change, however everywhere we call that function (only in frontend and clifton) we never explicitly state limit/page arguments. So I'm happy with pulling a fast one on this.


BPScott:
Build failure looks like a false negative - docker borked rather than the test suite.

votemike:
Test me please

PR #229, opened 2018-02-22
Adding countStreamableDescendantClips method

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6230


BPScott:
We don't need this as the data is already denormed onto the ProgrammeContainers and Episodes as `$programme->getAvailableClipsCount()`

PR #226, opened 2018-01-29
PROGRAMMES-6141 Allow counting of Episodes

votemike:
Required as part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6141

PR #230, opened 2018-02-22
Remove count queries for available episodes and clips

BPScott:
Revert #229 and #226, as the data provided by those queries is already present on the domain objects thanks to denormalisations.

* Use `$programme->getAvailableEpisodesCount()` instead of `ProgrammesAggregationService->countStreamableOnDemandEpisodes($programme)` (only on ProgrammeContainers)
* Use `$programme->getAvailableClipsCount()` instead of `ProgrammesAggregationService-> countStreamableDescendantClips($programme)` (on ProgrammeContainers and Episodes)

See https://github.com/bbc/programmes-frontend/pull/438 for removing the calls to these functions.


src/Service/ProgrammesAggregationService.php
```diff
@@ -5,6 +5,7 @@
 use BBC\ProgrammesCachingLibrary\CacheInterface;
 use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ClipRepository;
 use BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\CoreEntityRepository;
+use BBC\ProgrammesPagesService\Domain\ApplicationTime;
```
votemike: Is this used?

BPScott: No but I ran `git revert` on the commits that added this functionality, which included this bit of tidy up. Have added an extra commit to remove this again.

PR #231, opened 2018-02-26
Calculate isExternal within the RelatedLink domain object

BPScott:
The external field from PIPs doesn't seem to be used as diligently as we
hoped. Instead work out if a Related Link is external or not on our own.

A link is considered external if the url is on bbc.co.uk or bbc.com or
one of their subdomains.

This is done lazily so that we can try and avoid regex calls to parse
the url if they are not needed.

Add getHost to RelatedLink.

---

getHost is needed as related link lists in v2 shows the host if the link is external

### DIFF
--- a/src/Domain/Entity/RelatedLink.php
+++ b/src/Domain/Entity/RelatedLink.php
@@ -70,11 +70,6 @@ class RelatedLink
             $this->setUriMetadata();
         }
 
-        // Assert to make PHPStan confident that setUriMetadata sets isExternal
-        // to a not-null value. Without this PHPStan raises a wanring
-        // See https://github.com/phpstan/phpstan/issues/856
-        assert($this->isExternal !== null);
-
         return $this->isExternal;
     }
 
@@ -84,25 +79,21 @@ class RelatedLink
             $this->setUriMetadata();
         }
 
-        // Assert to make PHPStan confident that setUriMetadata sets host
-        // to a not-null value. Without this PHPStan raises a wanring
-        // See https://github.com/phpstan/phpstan/issues/856
-        assert($this->host !== null);
-
         return $this->host;
     }
 
     private function setUriMetadata(): void
     {
         $matches = [];
-        preg_match('@^(?:http[s]*://)?([^/?]+)@i', $this->uri, $matches);
+        preg_match('@^(?:https?://)([^/?]+)@i', $this->uri, $matches);
         $this->host = $matches[1] ?? '';
 
-        // A link is external if the hostname does not end with 'bbc.co.uk' or 'bbc.com'
+        // A link is external if the hostname is not empty and does not end with 'bbc.co.uk' or 'bbc.com'
         // Check the lengths, as strpos raises a warning if you try and use it
         // with a string shorter than the needle you're looking for
         $hostLength = strlen($this->host);
         $this->isExternal = !(
+            ($hostLength == 0) ||
             ($hostLength >= 9 && (strpos($this->host, 'bbc.co.uk', -9) !== false)) ||
             ($hostLength >= 7 && (strpos($this->host, 'bbc.com', -7) !== false))
         );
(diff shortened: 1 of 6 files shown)


src/Domain/Entity/RelatedLink.php
```diff
@@ -65,6 +66,45 @@ public function getType(): string
...
+
+        // Assert to make PHPStan confident that setUriMetadata sets isExternal
+        // to a not-null value. Without this PHPStan raises a wanring
+        // See https://github.com/phpstan/phpstan/issues/856
+        assert($this->isExternal !== null);
+
         return $this->isExternal;
     }
+
+    public function getHost(): string
```
votemike: Why do we need this method?

tests/Domain/Entity/RelatedLinkTest.php
```diff
@@ -24,5 +23,43 @@ public function testConstructor()
...
+            ['http://bbc.com/foo', 'bbc.com', false],
+            ['http://bbc.com?foo', 'bbc.com', false],
+
+            // Subdomains of bbc.co.uk / bbc.com count as internal links
+            ['http://www.bbc.co.uk/foo', 'www.bbc.co.uk', false],
+            ['http://brap.bbc.co.uk/foo', 'brap.bbc.co.uk', false],
+
+            // bbc.co.uk / bbc.com must be appear at the end
+            ['http://www.bbc.co.uk.nope/foo', 'www.bbc.co.uk.nope', true],
+            ['http://brap.bbc.co.uk.fake/foo', 'brap.bbc.co.uk.fake', true],
```
votemike: How about something like 'http://example.com?http://bbc.com'

src/Domain/Entity/RelatedLink.php
```diff
@@ -65,6 +66,45 @@ public function getType(): string
...
     public function isExternal(): bool
     {
+        if ($this->isExternal === null) {
+            $this->setUriMetadata();
+        }
+
+        // Assert to make PHPStan confident that setUriMetadata sets isExternal
+        // to a not-null value. Without this PHPStan raises a wanring
+        // See https://github.com/phpstan/phpstan/issues/856
+        assert($this->isExternal !== null);
```
BjornTwachtmann: I really don't like having code in here just to please PHPStan. Can't you just 
uUse the magic file to make it shut up instead

votemike: Why don't you set `$this->isExternal` in this method? And just get the other method to return it for you.

src/Domain/Entity/RelatedLink.php
BPScott: Because in ds2013 layouts it shows the hostname if the link is external (e.g. the related links on the promos at http://www.bbc.co.uk/programmes/p010772m)

src/Domain/Entity/RelatedLink.php
BPScott: @votemike `setUriMetadata()` sets both `$this->isExternal` and `$this->host`. Inlining the contents of `setUriMetadata()` into both `getHost()` and `isExternal()` works but the function is just big enough to make that duplication look weird. Having `setUriMetadata()` return an array of two values and doing `list($this->host, $this->isExternal) = $this->setUriMetadata();` in `getHost()` and `isExternal()` also does not work.

@BjornTwachtmann yep sure.

votemike: I meant something like this:
```
public function isExternal(): bool
{
    if ($this->isExternal === null) {
        $hostLength = strlen($this->getHost());
        $this->isExternal = !(
            ($hostLength == 0) ||
            ($hostLength >= 9 && (strpos($this->host, 'bbc.co.uk', -9) !== false)) ||
            ($hostLength >= 7 && (strpos($this->host, 'bbc.com', -7) !== false))
        );
    }
    return $this->isExternal;
}

public function getHost(): string
{
    if ($this->host === null) {
        $matches = [];
        preg_match('@^(?:https?://)([^/?]+)@i', $this->uri, $matches);
        $this->host = $matches[1] ?? '';
    }
    return $this->host;
}
```

This means that you won't need PHPStan magic.
I also think it's clearer that the other way.

PR #233, opened 2018-03-14
Adding media set mapping to domain entity

felipeparaujo:
Fetching the downloadable version is not possible in findByPidFull when the episode gets first fetched in programmes-frontend because we don't know what type of CoreEntity we're fetching, so Doctrine doesn't fetch that column. Any ideas on how to fetch it?

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
@@ -99,7 +99,7 @@ class VersionRepository extends EntityRepository
             ->addSelect(['versionTypes'])
             ->leftJoin('version.versionTypes', 'versionTypes')
             ->andWhere('version.programmeItem = :dbId')
-            ->andWhere('(version.streamable = 1 OR version.downloadable = 1)')
+            ->andWhere('version.streamable = 1 OR version.downloadable = 1')
             ->addOrderBy('version.pid', 'ASC')
             ->setParameter('dbId', $programmeDbId);
 
--- a/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
@@ -83,7 +83,7 @@ class VersionRepository extends EntityRepository
         return $qb->getQuery()->getResult(Query::HYDRATE_ARRAY)[0] ?? null;
     }
 
-    public function findStreamableByProgrammeItem(string $programmeDbId): array
+    public function findAvailableByProgrammeItem(string $programmeDbId): array
     {
         // YIKES! versionTypes is a many-to-many join, that could result in
         // an increase of rows returned by the DB and the potential for slow DB
@@ -99,7 +99,7 @@ class VersionRepository extends EntityRepository
             ->addSelect(['versionTypes'])
             ->leftJoin('version.versionTypes', 'versionTypes')
             ->andWhere('version.programmeItem = :dbId')
-            ->andWhere('version.streamable = 1')
+            ->andWhere('(version.streamable = 1 OR version.downloadable = 1)')
             ->addOrderBy('version.pid', 'ASC')
             ->setParameter('dbId', $programmeDbId);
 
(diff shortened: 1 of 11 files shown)


src/Domain/Entity/ProgrammeItem.php
```diff
@@ -29,6 +29,9 @@
     /** @var DateTimeImmutable|null */
     private $streamableUntil;
 
+    /** @var array */
+    private $downloadableMediaSets;
```
votemike: Is this just an array of arrays? Or an array of objects?

src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
```diff
@@ -279,7 +279,8 @@ private function getEpisodeModel(array $dbProgramme): Episode
             $dbProgramme['releaseDate'],
             $dbProgramme['duration'] ?? null,
             $this->castDateTime($dbProgramme['streamableFrom']),
-            $this->castDateTime($dbProgramme['streamableUntil'])
+            $this->castDateTime($dbProgramme['streamableUntil']),
+            $dbProgramme['downloadableMediaSets'] ?? []
```
BjornTwachtmann: Do you need to do an explode() here or something? I think you'll just get a comma separated string otherwise.

BPScott:
Out of interest why do we need the list of downloadable media sets here?

BjornTwachtmann:
@BPScott , it's for the download button, we need to know the list of media sets. Also, isn't it the middle of the night there ?!

src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
felipeparaujo: Doctrine takes care of the exploding, I tested that this works.

felipeparaujo:
@BPScott as Bjorn said, for the download button. 

Also, this is an easy way to know if there's a podcast version without joining into the versions table.

Also, go to bed.

src/Mapper/ProgrammesDbToDomain/CoreEntityMapper.php
BjornTwachtmann: huh, fair enough

src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
```diff
@@ -99,7 +99,7 @@ public function findStreamableByProgrammeItem(string $programmeDbId): array
             ->addSelect(['versionTypes'])
             ->leftJoin('version.versionTypes', 'versionTypes')
             ->andWhere('version.programmeItem = :dbId')
-            ->andWhere('version.streamable = 1')
+            ->andWhere('(version.streamable = 1 OR version.downloadable = 1)')
```
votemike: What about `is_podcastable`?

BPScott:
Gotcha on the "is this downloadable", thats a nice trick (also I'm in Korea atm because reasons).

I'm pondering will there ever be a need to get the list of media sets in the app, should there be an isDownloadable boolean on something in the hierarchy rather than a mediaSets array, that'd match $isStreamable that already exists?

votemike:
Which Korea?

BjornTwachtmann:
Best korea, I sincerely hope?

BjornTwachtmann:
@BPScott we need the list of media sets because we use them on the frontend to generate the dropdown, see https://github.com/bbc/programmes/blob/develop/webapp/php/sharedlib/BBC/Programmes/ProgrammeItem.php#L569-L589 . Please give my regards to the dear leader.

felipeparaujo:
I could add a isDownloadable method to ProgrammeItem that just checks if there are any media sets available. How's that sound?

src/Domain/Entity/ProgrammeItem.php
```diff
@@ -135,4 +140,19 @@ public function isVideo(): bool
...
+    }
+
+    public function isDownloadable(): bool
+    {
+        return !empty($this->downloadableMediaSets);
+    }
+
+    public function hasFutureAvailability(): bool
+    {
+        return !$this->isStreamable() && $this->getStreamableFrom();
```
votemike: What if the getStreamableFrom() is in the past?

felipeparaujo: Just added a comment explaining that.

PR #234, opened 2018-03-19
Fetching masterbrand, network and parents

votemike:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6215

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -423,8 +423,10 @@ QUERY;
         $filterOperation = $isNext ? '>' : '<' ;
 
         $qb = $this->getEntityManager()->createQueryBuilder()
-            ->select(['programme'])
+            ->select(['programme', 'masterbrand', 'network'])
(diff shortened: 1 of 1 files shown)


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -433,7 +435,13 @@ public function findAdjacentProgrammeByPosition(
             ->setParameter('parentDbId', $parentDbId)
             ->setParameter('originalPosition', $position);
 
-        return $qb->getQuery()->getOneOrNullResult(Query::HYDRATE_ARRAY);
+        $result = $qb->getQuery()->getOneOrNullResult(Query::HYDRATE_ARRAY);
+
+        if (!$result) {
+            return $result;
+        }
+        $this->addToAncestryCache([$result]);
```
BjornTwachtmann: Please read the comment on that method. You don't have all the joins on findById, so this will cause problems.

BjornTwachtmann: Best to remove it, not needed here anyway as we will already have the ancestors.

PR #235, opened 2018-03-21
PROGRAMMES-6261 - Find by Programme Item using Original Version

felipeparaujo:


src/Data/ProgrammesDb/EntityRepository/SegmentEventRepository.php
```diff
@@ -71,6 +71,32 @@ public function findByVersionWithContributions(array $dbIds, ?int $limit, int $o
         return $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
     }
 
+    public function findByProgrammeItemUsingOriginalVersion(int $dbId, ?int $limit, int $offset): array
```
votemike: Do you need the "By" in "findByProgrammeItemUsingOriginalVersion"?

src/Service/SegmentEventsService.php
```diff
@@ -94,6 +95,29 @@ function () use ($version, $limit, $page) {
         );
     }
 
+    public function findByProgrammeItemUsingOriginalVersion(
```
votemike: I'm a bit confused about this method. What's happening? What's being passed in and what is being returned?

felipeparaujo: Don't even bother. After talking to Bjorn, we've decided that just fetching all of the versions will be much easier.

PR #236, opened 2018-03-22
Find Collapsed Broadcasts by Programme

votemike:
As part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6276

Example viewable here: https://www.bbc.co.uk/programmes/b0082trb

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
@@ -168,21 +168,27 @@ QUERY;
     }
 
     public function findByProgramme(
-        array $ancestry,
+        int $programmeId,
         bool $isWebcastOnly,
         ?int $limit,
         int $offset
     ): array {
-        $qb = $this->createCollapsedBroadcastsOfProgrammeQueryBuilder($ancestry, $isWebcastOnly)
+        $qb = $this->createQueryBuilder('collapsedBroadcast', false)
+            ->addSelect(['programmeItem', 'image', 'masterBrand', 'mbNetwork'])
+            ->leftJoin('programmeItem.image', 'image')
+            ->leftJoin('programmeItem.masterBrand', 'masterBrand')
+            ->leftJoin('masterBrand.network', 'mbNetwork')
+            ->andWhere('collapsedBroadcast.isWebcastOnly = :isWebcastOnly')
+            ->andWhere('programmeItem.id = :programmeItemId')
+            ->setParameter('programmeItemId', $programmeId)
+            ->setParameter('isWebcastOnly', $isWebcastOnly)
             ->addOrderBy('collapsedBroadcast.endAt', 'ASC')
             ->setFirstResult($offset)
             ->setMaxResults($limit);
 
         $result = $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
 
-        $result = $this->explodeFields($result);
-
-        return $this->resolveProgrammeParents($result);
+        return $this->explodeFields($result);
     }
 
     public function findByProgrammeAndMonth(
--- a/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
@@ -167,6 +167,24 @@ QUERY;
         return $this->resolveProgrammeParents($result);
     }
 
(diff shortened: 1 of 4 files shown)


src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
```diff
@@ -167,6 +167,24 @@ public function findByCategoryAncestryAndStartAtDateRange(
         return $this->resolveProgrammeParents($result);
     }
 
+    public function findByProgramme(
+        array $ancestry,
+        bool $isWebcastOnly,
+        ?int $limit,
+        int $offset
+    ): array {
+        $qb = $this->createCollapsedBroadcastsOfProgrammeQueryBuilder($ancestry, $isWebcastOnly)
```
felipeparaujo: If you're going to use this just for showing the broadcasts in the episode page, you probably don't need all of the joins that this method adds. Can you make this query slimmer?

src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
```diff
@@ -167,6 +167,24 @@ public function findByCategoryAncestryAndStartAtDateRange(
...
+            ->addOrderBy('collapsedBroadcast.endAt', 'ASC')
+            ->setFirstResult($offset)
+            ->setMaxResults($limit);
+
+        $result = $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
+
+        $result = $this->explodeFields($result);
+
+        return $this->resolveProgrammeParents($result);
```
felipeparaujo: Same thing: do you actually need the whole ancestry resolved?

BjornTwachtmann: The ancestry should already be cached in memory at this point, as it will inevitably consist of the episode this is a broadcast of and its parents, which will already have been queried.

src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
BjornTwachtmann: Yeah, I don't think createCollapsedBroadcastsOfProgrammeQueryBuilder is right here. You are looking for broadcasts of a single episode, that will give you broadcasts of an episode and of all of its children. Not that the episode will have children, but that's a much slower query. We just need to ascertain that the collapsed broadcast has a programme item ID the same as our programme. 

I would probably add the same joins as that thing does though, chances are some of our presenters will be expecting them and relational databases are exceptionally fast at doing joins if you're not trying to filter by both sides of them, which we aren't.

src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
```diff
@@ -167,6 +167,30 @@ public function findByCategoryAncestryAndStartAtDateRange(
...
+            ->andWhere('programmeItem.id = :programmeItemId')
+            ->setParameter('programmeItemId', $programmeId)
+            ->setParameter('isWebcastOnly', $isWebcastOnly)
+            ->addOrderBy('collapsedBroadcast.endAt', 'ASC')
+            ->setFirstResult($offset)
+            ->setMaxResults($limit);
+
+        $result = $qb->getQuery()->getResult(Query::HYDRATE_ARRAY);
+
+        return $this->explodeFields($result);
```
BjornTwachtmann: I think you should probably still be resolving programme parents here, unless you're very sure you don't need them

votemike: I don't need them. I don't know if someone will in the future.

BPScott: If you don't resolve the parents then things like image and masterbrand inheritance on the ProgrammeItem will not work correctly (as it won't have a parent to check for an image's existence).

If you don't need the ProgrammeItem that is attached to the CollapsedBroadcast at all (which sounds like it might be the case as you're already filtering by it), then don't bother joining to programmeItem (and thus), Image, MasterBrand and Network.

BjornTwachtmann: Looking at it again, it's probably also a good idea to have this sort by start date, just because there's already an index on that field, and it should make bugger all difference to the output.

PR #239, opened 2018-03-28
PROGRAMMES-6279 collections

votemike:
Needed for https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6279

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -132,7 +132,7 @@ class CoreEntityRepository extends MaterializedPathRepository
         return $this->resolveParents($result);
     }
 
-    public function findByCoreEntityMembership(int $entityId, string $groupType = 'Group', ?int $limit, int $offset): array
+    public function findByCoreEntityMembership(int $entityId, string $groupType, ?int $limit, int $offset): array
     {
         $this->assertEntityType($groupType, ['Group', 'Collection', 'Franchise', 'Gallery', 'Season']);
 
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -132,17 +132,18 @@ class CoreEntityRepository extends MaterializedPathRepository
         return $this->resolveParents($result);
     }
 
-    public function findByCoreEntityMembership(int $entityId, ?string $groupType, ?int $limit, int $offset)
+    public function findByCoreEntityMembership(int $entityId, string $groupType = 'Group', ?int $limit, int $offset): array
     {
-        if ($groupType === null) {
-            $groupType = 'Group';
-        }
+        $this->assertEntityType($groupType, ['Group', 'Collection', 'Franchise', 'Gallery', 'Season']);
 
         $qb = $this->getEntityManager()->createQueryBuilder()
-            ->select(['groupEntity', 'image'])
+            ->select(['groupEntity', 'image', 'masterBrand', 'network', 'mbImage'])
             ->from('ProgrammesPagesService:' . $groupType, 'groupEntity')
             ->join('ProgrammesPagesService:Membership', 'membership', Query\Expr\Join::WITH, 'membership.group = groupEntity')
             ->leftJoin('groupEntity.image', 'image')
+            ->leftJoin('groupEntity.masterBrand', 'masterBrand')
+            ->leftJoin('masterBrand.network', 'network')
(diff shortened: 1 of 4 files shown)


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -132,6 +132,28 @@ public function findAvailableEpisodesByCategoryAncestry(
         return $this->resolveParents($result);
     }
 
+    public function findByCoreEntityMembership(int $entityId, ?string $groupType, ?int $limit, int $offset)
+    {
+        if ($groupType === null) {
```
BjornTwachtmann: Can you do this the same way as the other stuff in the class, make $groupType default to 'Group' and validate what's passed in against the constant of valid entity types. Realise that's not completely validating it's a group, but it should protect against anything truly messy.

BjornTwachtmann:
Aside from that one comment, looks good. I'm not sure if we need to be joining to masterbrand etc. in that query, but it should probably work without

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -132,6 +132,28 @@ public function findAvailableEpisodesByCategoryAncestry(
...
+            $groupType = 'Group';
+        }
+
+        $qb = $this->getEntityManager()->createQueryBuilder()
+            ->select(['groupEntity', 'image'])
+            ->from('ProgrammesPagesService:' . $groupType, 'groupEntity')
+            ->join('ProgrammesPagesService:Membership', 'membership', Query\Expr\Join::WITH, 'membership.group = groupEntity')
+            ->leftJoin('groupEntity.image', 'image')
```
BjornTwachtmann: Thinking about it again, please join to master brand, network and network image here, the domain object construction expects these things to be present and we don't want to be chasing obscure bugs for the sake of extra microseconds on the query,

src/Service/GroupsService.php
```diff
@@ -33,6 +37,59 @@ public function __construct(
...
+                'Called %s with an invalid type. Expected a Programme, Collection or Gallery but got "%s"',
+                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'],
+                \get_class($coreEntity)
+            ));
+        }
+
+        if (!\in_array($groupType, ['Group', 'Collection', 'Franchise', 'Gallery', 'Season'])) {
+            throw new InvalidArgumentException(sprintf(
+                'Called %s with an invalid type. Expected Group, Collection, Franchise, Gallery or Season but got "%s"',
+                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'],
```
BPScott: This debug_backtrace weirdness is because in other class this check is extracted out into a private method, but we want the exception message to say about "Called findByCoreEntityMembership with an invalid type".

If you're doing the throw the actual method instead of a helper method you can use `__FUNCTION__` to get the name of the current function instead of parsing the backtrace to find the calling function.

src/Service/GroupsService.php
```diff
@@ -33,6 +37,59 @@ public function __construct(
...
+     */
+    public function findByCoreEntityMembership(
+        CoreEntity $coreEntity,
+        string $groupType = 'Group',
+        ?int $limit = self::DEFAULT_LIMIT,
+        int $page = self::DEFAULT_PAGE,
+        $ttl = CacheInterface::NORMAL,
+        $nullTtl = CacheInterface::NORMAL
+    ): array {
+        if (!($coreEntity instanceof Programme || $coreEntity instanceof Collection || $coreEntity instanceof Gallery)) {
```
BPScott: Any Group /could/ be a member of another Group, I don't think it's worth having this check here.

src/Service/GroupsService.php
```diff
@@ -33,6 +37,59 @@ public function __construct(
...
+                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'],
+                \get_class($coreEntity)
+            ));
+        }
+
+        if (!\in_array($groupType, ['Group', 'Collection', 'Franchise', 'Gallery', 'Season'])) {
+            throw new InvalidArgumentException(sprintf(
+                'Called %s with an invalid type. Expected Group, Collection, Franchise, Gallery or Season but got "%s"',
+                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'],
+                \get_class($coreEntity)
```
BPScott: This should be `$groupType`. Currently you're printing out the error message for an invalid second argument, but you're mentioning the value of the first argument

PR #242, opened 2018-03-29
Acting on PR comments

votemike:
Acting on PR review from a different timezone: https://github.com/bbc/programmes-pages-service/pull/239#pullrequestreview-107797728

### DIFF
--- a/src/Service/GroupsService.php
+++ b/src/Service/GroupsService.php
@@ -54,19 +54,10 @@ class GroupsService extends AbstractService
         $ttl = CacheInterface::NORMAL,
         $nullTtl = CacheInterface::NORMAL
     ): array {
-        if (!($coreEntity instanceof Programme || $coreEntity instanceof Collection || $coreEntity instanceof Gallery)) {
-            throw new InvalidArgumentException(sprintf(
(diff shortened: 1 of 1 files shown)


src/Service/GroupsService.php
```diff
@@ -54,19 +54,10 @@ public function findByCoreEntityMembership(
...
-            ));
-        }
-
         if (!\in_array($groupType, ['Group', 'Collection', 'Franchise', 'Gallery', 'Season'])) {
             throw new InvalidArgumentException(sprintf(
-                'Called %s with an invalid type. Expected Group, Collection, Franchise, Gallery or Season but got "%s"',
-                debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'],
-                \get_class($coreEntity)
+                'Called findByCoreEntityMembership with an invalid groupType. Expected Group, Collection, Franchise, Gallery or Season but got "%s"',
+                \get_class($groupType)
```
BPScott: Better, but now it'll just say "...but got string", it'd  be useful to say what the value of the string was :D so `$groupType` rather than `\get_class($groupType)`

PR #243, opened 2018-03-29
Optimize imports

votemike:


src/Data/ProgrammesDb/Entity/AtozTitle.php
```diff
@@ -2,8 +2,6 @@
 
 namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
 
-use Doctrine\ORM\Mapping as ORM;
```
votemike: This was in a load of files. Is it needed for the annotations or something?

BjornTwachtmann:
@votemike removing that import will completely break doctrine, it's needed for the annotations. If you clear your symfony cache and test this you'll see everything's borked.

votemike:
Another reason that comments shouldn't be used as code.
I will fix.

PR #249, opened 2018-06-25
Programmes 6349 smp playlist

BjornTwachtmann:
* Add isEmbeddable to ProgrammeItem Domain object
* Add query to get all streamable versions with the canonical one first
* Add query to get a ProgrammeItem together with any version warnings


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -636,6 +631,36 @@ public function coreEntityAncestryGetter(array $ids): array
         return $cached;
     }
 
+    private function findByPidFullCommon(string $pid, string $entityType = 'CoreEntity'): QueryBuilder
```
votemike: Should `$entityType` have a default? Why not just make it required?

src/Domain/Entity/Episode.php
```diff
@@ -35,6 +35,7 @@ public function __construct(
         int $aggregatedBroadcastsCount,
         int $availableClipsCount,
         int $aggregatedGalleriesCount,
+        bool $isEmbeddable,
```
votemike: What does this mean?

src/Domain/Entity/Image.php
```diff
@@ -63,7 +65,12 @@ public function getLongestSynopsis(): string
     public function getUrl($width, $height = 'n'): string
     {
         $recipe = $width . 'x' . $height;
-        return 'https://ichef.bbci.co.uk/images/ic/' . $recipe . '/' . $this->filename;
+        return self::BASE_URL . '/' . $recipe . '/' . $this->filename;
+    }
+
+    public function getRecipeUrl(): string
+    {
+        return self::BASE_URL . '/$recipe/' . $this->filename;
```
votemike: This looks similar but different to the one above. Is it definitely correct?

src/Domain/Entity/Episode.php
BjornTwachtmann: Some clips can be embedded into other sites, this is controlled by a field in PIPs, which we need to expose in the domain object for the frontend PR to have the info

src/Domain/Entity/Image.php
BjornTwachtmann: It's very similar, I wasn't sure whether to put this here or not really, but I thought it's better to have this stuff defined in one place. 
Some stuff needs a URL where the dimensions can be dynamically relaced (SMP)

I could just make the base url constant public and add a getter for the filename otherwise. Any opinions?

src/Domain/Entity/Episode.php
votemike: Should this have a docblock? Or be changed to `$isExternallyEmeddable` or something?

src/Domain/Entity/Image.php
votemike: FML. Maybe a note that it is correct to be a string and isn't supposed to be a variable?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
BjornTwachtmann: Yeah, fair point.

src/Domain/Entity/Episode.php
BjornTwachtmann: there's about 100 parameters here, so a docblock may not be that helpful, but I can change the name if people think that's clearer

PR #250, opened 2018-06-28
Adding Podcast Entity

votemike:
Part of https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6433


src/Data/ProgrammesDb/Entity/CoreEntity.php
```diff
@@ -349,6 +349,16 @@ public function setImage(?Image $image): void
         $this->image = $image;
     }
 
+    public function getIsPodcastable(): bool
```
BjornTwachtmann: These are already defined on ProgrammeContainer and Collection. Shouldn't need them here. If anything adjust the query in Faucet so it only returns the relevant types of CoreEntity

PR #258, opened 2018-08-08
Add method to clear associated versions from a ProgrammeItem to aid i…

BjornTwachtmann:
…n a nasty Faucet hack


src/Data/ProgrammesDb/Entity/ProgrammeItem.php
```diff
@@ -224,4 +224,11 @@ public function setSegmentEventCount(int $segmentEventCount): void
     {
         $this->segmentEventCount = $segmentEventCount;
     }
+
+    public function clearVersionAssociations(): void
```
votemike: I assume this is only to be used in special circumstances. Should there be a warning comment on it?

PR #264, opened 2018-10-03
Fix embargoed competition warning bug

BjornTwachtmann:


tests/Mapper/ProgrammesDbToDomain/MasterBrandMapperTest.php
```diff
@@ -170,6 +170,35 @@ public function testGetDomainModelWithSetCompetitionWarning()
...
+    {
+        $networkDbEntity = ['nid' => 'bbc_one'];
+        $versionDbEntity = ['pid' => 'p01m5mss', 'programmeItem' => null];
+
+        $this->setupNetworkMapper($networkDbEntity, $this->mockNetwork);
+
+        $this->mockVersionMapper->expects($this->never())
+            ->method('getDomainModel');
+
+        $dbEntityArray = [
```
cisco87: Are you sure you won't have a cleaner test by putting these stuff in a dataProvider?

src/Mapper/ProgrammesDbToDomain/MasterBrandMapper.php
```diff
@@ -91,6 +91,12 @@ private function getCompetitionWarningModel(array $dbMasterBrand, string $key =
             return null;
         }
 
+        if (array_key_exists('programmeItem', $dbMasterBrand[$key]) && is_null($dbMasterBrand[$key]['programmeItem'])) {
```
votemike: So, it's fine if the programmeItem key doesn't exist?

BjornTwachtmann: This is basically the difference between "Not queried" and "Not found". Hence using array_key_exists instead of isset

PR #267, opened 2018-10-29
PROGRAMMES-6680: Make streamable versions come with competition warni…

BjornTwachtmann:
…ngs. Including for ProgrammeItems where only their parents have masterbrands

* This changes some things in the versionsService in order to make sure that competition warnings are retreived correctly and attached to a streamableVersion in all cases. Particularaly those cases where a ProgrammeItem has no masterbrand, but its parent does. 

Tidies up some things around making sure that the programmeItems attached to versions contain all the joins from the query.

It's a bit magic, so ask if you have questions.

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
@@ -240,7 +240,7 @@ class VersionRepository extends EntityRepository
      * @param array $versions
      * @return array $versions
      */
-    private function hydrateCompetitionWarnings(array $versions)
+    private function hydrateCompetitionWarnings(array $versions): array
     {
         $versionsToFetchFor = [];
 
@@ -270,7 +270,7 @@ class VersionRepository extends EntityRepository
      * @param array $programmeItems
      * @return array $programmeItems
      */
-    private function hydrateProgrammeItems(array $programmeItems)
+    private function hydrateProgrammeItems(array $programmeItems): array
     {
         $streamableVersions = [];
         foreach ($programmeItems as $index => &$programmeArray) {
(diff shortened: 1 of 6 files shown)


src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
```diff
@@ -209,4 +222,86 @@ public function createQueryBuilder($alias, $indexBy = null)
...
+     *
+     * It ensures those things get the correct competition warning.
+     *
+     * In 90% of cases, this will just return the input array.
+     *
+     * In the other cases it queries all the programme's parents, and their masterbrands
+     * joined to their competition warnings, adds them to the version, and returns that
+     * whole mess.
+     *
+     * @param array $versions
```
Tasiobg: I think this also make sense
`@param version[] $versions`

src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
```diff
@@ -209,4 +222,86 @@ public function createQueryBuilder($alias, $indexBy = null)
...
+     * @return array $versions
+     */
+    private function hydrateCompetitionWarnings(array $versions)
```
Tasiobg: You can add the returned type
```private function hydrateCompetitionWarnings(array $versions): array```

Also I'm assuming `$versions` can't be null

src/Data/ProgrammesDb/EntityRepository/VersionRepository.php
BjornTwachtmann: It's not though, it's the raw array that comes back from Doctrine, not an array of Version objects

PR #271, opened 2018-11-15
Programmes-6671: episodes downloads view

BjornTwachtmann:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6671

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
@@ -31,6 +31,8 @@ class CoreEntityRepository extends MaterializedPathRepository
         'Franchise',
     ];
 
+    const DB_GROUP_TYPES = ['collection', 'season', 'franchise', 'gallery'];
+
     private $ancestryCache = [];
 
     public function findTleosByCategory(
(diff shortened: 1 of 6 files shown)


src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -184,7 +186,36 @@ public function findByPid(string $pid, string $entityType = 'CoreEntity'): ?arra
      */
     public function findByPidFull(string $pid, string $entityType = 'CoreEntity'): ?array
     {
-        $qb = $this->findByPidFullCommon($pid, $entityType);
+        $qb = $this->findFullCommon($entityType);
+        $qb->andWhere('entity.pid = :pid')
```
cisco87: I prefer the $qb->expr()->eq(), if we have to use the QueryBuilder I think we should fully embrace it, don't you think?

src/Data/ProgrammesDb/EntityRepository/CoreEntityRepository.php
```diff
@@ -184,7 +186,36 @@ public function findByPid(string $pid, string $entityType = 'CoreEntity'): ?arra
...
+
+    /**
+     * Full Find By Pid
+     *
+     * This resolves all parents
+     */
+    public function findByIdFull(int $dbId, string $entityType = 'CoreEntity'): ?array
+    {
+        $qb = $this->findFullCommon($entityType);
+        $qb->andWhere('entity.id = :id')
```
cisco87: Same $qb->expr()->eq

PR #272, opened 2018-11-15
PROGRAMMES-6707 - Count method

cisco87:
Needed by https://github.com/bbc/programmes-frontend/pull/810

### DIFF
--- a/src/Service/ProgrammesAggregationService.php
+++ b/src/Service/ProgrammesAggregationService.php
@@ -47,7 +47,7 @@ class ProgrammesAggregationService extends AbstractService
         $ttl = CacheInterface::NORMAL,
         $nullTtl = CacheInterface::NORMAL
     ) : int {
-        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $programme->getPid(), 'ClipsCount', null, null, $ttl);
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $programme->getPid());
 
         return $this->cache->getOrSet(
             $key,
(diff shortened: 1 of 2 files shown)


src/Service/ProgrammesAggregationService.php
```diff
@@ -42,6 +42,24 @@ public function findStreamableDescendantClips(
         return $this->findStreamableDescendantsByType($programme, 'Clip', $limit, $page, $ttl, $nullTtl);
     }
 
+    public function countStreamableDescendantClips(
+        Programme $programme,
+        $ttl = CacheInterface::NORMAL,
+        $nullTtl = CacheInterface::NORMAL
+    ) : int {
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $programme->getPid(), 'ClipsCount', null, null, $ttl);
```
BjornTwachtmann: You don't need the two null parameters here, or that clipscount string

cisco87: Sorry about the 2 nulls, but I tought the ClipsCount was a sort of prefix to make the cache key unique?

cisco87: ok I understand now, it's the __FUNCTION__ that will help generate it, got it!

src/Service/ProgrammesAggregationService.php
```diff
@@ -42,6 +42,24 @@ public function findStreamableDescendantClips(
         return $this->findStreamableDescendantsByType($programme, 'Clip', $limit, $page, $ttl, $nullTtl);
     }
 
+    public function countStreamableDescendantClips(
+        Programme $programme,
+        $ttl = CacheInterface::NORMAL,
+        $nullTtl = CacheInterface::NORMAL
+    ) : int {
+        $key = $this->cache->keyHelper(__CLASS__, __FUNCTION__, $programme->getPid());
```
BjornTwachtmann: You do, however, need the TTL in there

PR #276, opened 2019-01-04
PROGRAMMES-6754 - hide barkers and off-airs

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6754

### DIFF
--- a/src/Data/ProgrammesDb/Entity/Programme.php
+++ b/src/Data/ProgrammesDb/Entity/Programme.php
@@ -36,6 +36,13 @@ abstract class Programme extends CoreEntity implements GalleriesCountableInterfa
      */
     private $firstBroadcastDate;
 
+    /**
+     * @var bool
+     *
+     * @ORM\Column(type="boolean", nullable=false, options={"default" = 0})
+     */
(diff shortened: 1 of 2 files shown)


Bobkidbob:
Test this please.

src/Data/ProgrammesDb/Entity/Programme.php
```diff
@@ -129,4 +136,14 @@ public function setFirstBroadcastDate(?DateTime $firstBroadcastDate): void
     {
         $this->firstBroadcastDate = $firstBroadcastDate;
     }
+
+    public function getIsGap(): bool
```
cisco87: AFAIR (Remember :D) doctrine also tries (isGap()) ? Which I think might be better? No?

BjornTwachtmann: We use getIs...() everywhere else, Doctrine isn't doing any magic here, it's just a getter

cisco87: Alright got it, then better to make it consistent, sorry

PR #282, opened 2019-04-04
Fix stopwords <= 2 characters breaking searches

BjornTwachtmann:
If you visit something like /programmes/a-z/by/i am doctor who/all on test, you'll see no results. This is because MySQL treats anything <= 2 characters as a stopword, and in a boolean query will return no results.

This PR removes words <= 2 characters from terms actually searched for, and improves handling of search input slightly. 

https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6937?focusedCommentId=4261818&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4261818


tests/Data/ProgrammesDb/Util/SearchUtilitiesTraitTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public function setUp()
+    {
+        $this->searchUtilityObject = $this->getObjectForTrait(SearchUtilitiesTrait::class);
+    }
+
+    /**
+     * @dataProvider stripPunctuationDataProvider
+     */
+    public function testStripPunctuation(string $string, string $expectedString)
```
Tasiobg: Looks like you could add the returning types to the functions in this class

src/Data/ProgrammesDb/Util/SearchUtilitiesTrait.php
```diff
@@ -0,0 +1,33 @@
...
+        return $string ?? '';
+    }
+
+    protected function makeBooleanSearchQuery(string $keywords): ?string
+    {
+        $keywords = $this->stripPunctuation($keywords);
+        // Split into words
+        $words = preg_split('/\s+/', $keywords, 20, PREG_SPLIT_NO_EMPTY);
+        // Filter to words >= 2 chars
+        $validWords = array_filter($words, function ($word) {
```
Tasiobg: I learn something interesting about anonymous functions, you can declare them also as `static` so they don't get bound to the current class but this is a microptimization, doesn't really make a difference

src/Data/ProgrammesDb/Util/SearchUtilitiesTrait.php
```diff
@@ -0,0 +1,33 @@
...
+            return (strlen($word) > 2);
+        });
+        if (!empty($validWords)) {
+            // Join into MySQL compatible boolean query string
+            $booleanKeywords = join(' +', $validWords);
```
Tasiobg:
```suggestion
            $booleanKeywords = implode(' +', $validWords);
```
join is an alias of implode, does it matters?

tests/Data/ProgrammesDb/Util/SearchUtilitiesTraitTest.php
BjornTwachtmann: You mean that expectedString should be nullable? Or which return types are we talking about here?

Tasiobg: I mean this
```suggestion
    public function testStripPunctuation(string $string, string $expectedString): void
```

BjornTwachtmann: ah, okay, we don't generally stick return type hints on our unit tests, I hadn't considered it. You think it's worth starting to do it?

src/Data/ProgrammesDb/Util/SearchUtilitiesTrait.php
BjornTwachtmann: Interesting, didn't know you could do that. Can't see it making much difference here though.

src/Data/ProgrammesDb/Util/SearchUtilitiesTrait.php
BjornTwachtmann: Implode does sound cooler

tests/Data/ProgrammesDb/Util/SearchUtilitiesTraitTest.php
Tasiobg: Could be a good practice for unitest as well but not something necesary because it has very little impact, so whatever you fancy

PR #286, opened 2019-04-25
PROGRAMMES-6731 - get top level genres and their children

Bobkidbob:
https://jira.dev.bbc.co.uk/browse/PROGRAMMES-6731

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
@@ -98,6 +98,7 @@ class CategoryRepository extends MaterializedPathRepository
             ->andWhere('category INSTANCE OF ProgrammesPagesService:Genre')
             ->andWhere('category.depth = 1')
             ->addOrderBy('category.urlKey')
+            ->addOrderBy('children.urlKey')
             ->getQuery()->getResult(Query::HYDRATE_ARRAY);
 
         return $this->resolveParents($result);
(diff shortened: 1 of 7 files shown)


src/Domain/Entity/Category.php
```diff
@@ -73,6 +73,8 @@ public function getDbAncestryIds(): array
 
     abstract protected function getAncestry(): array;
 
+    abstract protected function getChildren(): array;
```
BjornTwachtmann: Should be public, I change the other ones in the PR that I'm working on.

src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
```diff
@@ -90,6 +90,20 @@ public function findAllByTypeAndMaxDepth(string $type, int $maxDepth): array
         return $this->resolveParents($result);
     }
 
+    public function findGenresAndChildren(): array
```
BjornTwachtmann: findTopLevelGenresAndChildren?

src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
```diff
@@ -90,6 +90,20 @@ public function findAllByTypeAndMaxDepth(string $type, int $maxDepth): array
...
+    {
+        $result = $this->createQueryBuilder('category')
+            ->addSelect('children')
+            ->innerJoin('category.children', 'children')
+            ->andWhere('category INSTANCE OF :type')
+            ->andWhere('category.depth = 1')
+            ->addOrderBy('category.urlKey')
+            ->setParameter('type', 'genre')
```
BjornTwachtmann: doesn't need to be a parameter

src/Domain/Entity/Genre.php
```diff
@@ -66,6 +71,11 @@ public function getAncestry(): array
         return $this->ancestry;
     }
 
+    public function getChildren(): ?array
```
BjornTwachtmann: This needs to throw an DataNotFetchedException if $this->children is null rather than return null

PR #289, opened 2019-07-10
PROGRAMMES-7087: Queries for season page

BjornTwachtmann:
This isn't going to be the easiest PR to understand. 

Performance of these queries without making major changes to data structures requires some fairly funky SQL that doctrine won't implement in DQL, hence native queries and some slightly unusual structures. 

Probably easiest to just ask if you want this explained.


Tasiobg:
I had a look at the code and seems sensible but I'm not very familiar with doctrine and not as familiar as you with the DB and the data relations so feel free to merge it when you want.

Build is failing, is that on purpose?

BjornTwachtmann:
@Tasiobg obviously I failed the build on purpose :) . It was just PHPCS complaining, I fixed it.

PR #294, opened 2019-08-12
Prevent duplicate categories from blowing up doctrine with a limit statement

BjornTwachtmann:
PROGRAMMES-7112

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
@@ -41,7 +41,8 @@ class CategoryRepository extends MaterializedPathRepository
         $finalI = count($urlKeys);
         $query->addSelect('category' . $finalI)
             ->leftJoin('category' . ($finalI - 1) . '.parent', 'category' . $finalI)
-            ->andWhere('category' . $finalI . '.id IS NULL');
+            ->andWhere('category' . $finalI . '.id IS NULL')
+            ->setMaxResults(1);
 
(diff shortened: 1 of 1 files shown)


src/Data/ProgrammesDb/EntityRepository/CategoryRepository.php
```diff
@@ -41,7 +41,8 @@ public function findByUrlKeyAncestryAndType(array $urlKeys, string $type): ?arra
         $finalI = count($urlKeys);
         $query->addSelect('category' . $finalI)
             ->leftJoin('category' . ($finalI - 1) . '.parent', 'category' . $finalI)
-            ->andWhere('category' . $finalI . '.id IS NULL');
+            ->andWhere('category' . $finalI . '.id IS NULL')
+            ->setMaxResults(1);
```
gmarinov: will this ever end up selecting the "wrong" of the two (or more)?
thoughts on a unique index in the db?

BjornTwachtmann: It could end up choosing the "wrong" one, but in the case of this kind of duplicate we have no idea of what the right one is really, it's pretty arbitrary.

URL keys are not unique across the data set, but they should be unique within a given part of the data set (top level, or underneath a given format). We could enforce that programmatically in Faucet I suppose.

gmarinov: code lgtm. 
I'd vote for it excepting on crap data though (or at least signalling) so it can be fed back to whoever created the messy data..

PR #303, opened 2020-01-24
5996 - Brandingtool-specific functionality

gmarinov:
adds MasterBrand-related functionality

see https://github.com/bbc/brandingtool-cloud/pull/164

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,3 +141,131 @@ This repository is available under the terms of the Apache 2.0 license.
 View the [LICENSE file](LICENSE) for more information.
 
 Copyright (c) 2017 BBC
+
+
+New Yaml
+--------
+
+in services.yml
+
+```yaml
+
+framework:
+    cache:
+        pools:
+            cache.programmes:
+                adapter: 'cache.adapter.psr6'
+                provider: cache.null_provider
+services:
+    cache.null_provider:
+        class: Symfony\Component\Cache\Adapter\NullAdapter
+
+    BBC\ProgrammesCachingLibrary\Cache:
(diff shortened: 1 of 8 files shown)


src/Domain/Entity/Season.php
```diff
@@ -12,6 +13,12 @@ class Season extends Group
     /* @var int */
     private $aggregatedBroadcastsCount;
 
+    /* @var DateTime */
```
BjornTwachtmann: Just remembered this, but all the other domain models use DateTimeImmutable for date properties to prevent people accidentally/onpurpose changing things in here once it's instantiated. 

Would be better to keep that consistent I think, it's literally a case of replacing the class name here and in the mapper

BjornTwachtmann: Actually not even the mapper I think, just the typehint, doctrine seems to be already returning the right class.

README.md
```diff
@@ -141,3 +141,51 @@ This repository is available under the terms of the Apache 2.0 license.
 View the [LICENSE file](LICENSE) for more information.
 
 Copyright (c) 2017 BBC
+
+
+New Yaml
```
BjornTwachtmann: If we're adding the Yaml for how to bring this thing into Symfony it probably needs describing a bit, especially what it's doing with caching etc.

src/Domain/Entity/Season.php
gmarinov: there was a reason I ended up using `DateTime` but I can’t remember it now :) I’ll try again.

PR #308, opened 2024-07-23
DATCAP-1045: Retain Dazzler broadcasts for Collapsed Broadcast

Tyrannican:
This PR adds a check for the `stripWebcasts` function which removed all Webcast broadcasts from the Collapsed Broadcast Service. This was stripping out Dazzler broadcasts which should be retained so an additional check was added here to allow them to still be included.

I'll preface this with the following: `I have absolutely no idea if this will work but shouldn't cause any issues otherwise`

### DIFF
--- a/src/Service/CollapsedBroadcastsService.php
+++ b/src/Service/CollapsedBroadcastsService.php
@@ -815,7 +815,7 @@ class CollapsedBroadcastsService extends AbstractService
 
     private function isDazzlerBroadcast(mixed $serviceId): bool
     {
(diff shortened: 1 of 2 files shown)


Janelle-BBC (approved):

murkas (approved):
Clean solution @Tyrannican

kmurphy3377 (approved):

darceyh03 (approved):

Tyrannican:
Closing as another fix is planned

PR #309, opened 2024-07-25
DATCAP-1045: Filter webcast for `findUpcomingUnderGroup` (Dazzler) query

Tyrannican:
This PR adds a simpler fix for filtering the Dazzler broadcasts by editing the SQL query in which Dazzler broadcasts are gathered.

### DIFF
--- a/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
+++ b/src/Data/ProgrammesDb/EntityRepository/CollapsedBroadcastRepository.php
@@ -337,19 +337,19 @@ QUERY;
          *
          * The rest of the query is obvious and deals with getting the relevant metadata around the core entity etc.
          */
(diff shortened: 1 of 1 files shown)


colin-campbell-bbc (approved):

murkas (approved):

murkas (commented):
There's potentially an improvement to be had by doing both `INNER JOIN`s and then filtering with `WHERE` since the two results sets are agglomerated by the `UNION`, but this is a minor performance improvement - no need to think about this now if we don't desire to cause additional headaches!

Tyrannican:
Yeah, my SQL knowledge is rusty at best and don't want to risk breaking things :sweat_smile: 
can potentially look at it in the future though!

kmurphy3377 (approved):

PR #311, opened 2024-10-18
DATCAP-1031: Update Image Entity

kmurphy3377:
## Ticket
[DATCAP-1031](https://jira.dev.bbc.co.uk/browse/DATCAP-1031)

## Changes
- Added two new properties for Image entity: height and width
- Added getters and setters for the two new properties

## Impact
Brings the Image Entity into line with the database, now possessing a height and width property. Will be used in Faucet to allow image height and width to be ingested into the Programmes DB

## Testing
Unit tests are passing

### DIFF
--- a/src/Data/ProgrammesDb/Entity/Image.php
+++ b/src/Data/ProgrammesDb/Entity/Image.php
@@ -122,7 +122,7 @@ class Image implements MemberOfGroupInterface, RelatedLinkContextInterface, Prom
         $this->extension = $extension;
     }
 
-    public function getHeight(): int
+    public function getHeight(): ?int
     {
         return $this->height;
     }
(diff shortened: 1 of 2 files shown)


src/Data/ProgrammesDb/Entity/Image.php
```diff
@@ -107,4 +121,24 @@ public function setExtension(string $extension): void
     {
         $this->extension = $extension;
     }
+
+    public function getHeight(): ?int
+    {
+        return $this->height;
+    }
+
+    public function setHeight(int $height): void
```
colin-campbell-bbc, line 130: For setting height and width, should the params be nullable integers? I'm not sure if PIPs always includes the dimensions

kmurphy3377: Ahh yes they should! Good catch! Updated now 😄

murkas (approved):

ducampbe (approved):

PR #312, opened 2024-10-25
Add type & getType

colin-campbell-bbc:
The DB entity needs to be able to return what type it is so that podcasts belonging to Brand, Series or Collection are the only ones that can be ingested.


RachelODonnellBBC (approved):

colin-campbell-bbc:
Closing because the build is broken in _many_ places

PR #314, opened 2025-01-29
DATCAP-1205: Create Ancestry Entity

murkas:
Ticket
-----------------
https://jira.dev.bbc.co.uk/browse/DATCAP-1205

Changes
----------------
Created an Ancestry entity.
Created a test file for the Ancestry entity.
Created an Ancestry domain file to establish the ancestry domain.
Created an AncestryRepository file (currently empty), to store ancestry specific queries.
Created an AncestryService file, to allow ancestry instances to be served from the database.
Created an AncestryMapper file to connect the ancestry service to the ancestry entity.
Added the AncestryMapper to the file of avaliable mappers.

Impact
---------------
Ancestry table created in schema. No changes to coreEntity table.

Testing
--------------
The programmes-pages-service build is passing: https://eu-west-1.console.aws.amazon.com/codesuite/codebuild/288896551516/projects/programmes-pages-service/build/programmes-pages-service%3A922e83a0-fe62-4aba-bfa7-e3fe980ff73e?region=eu-west-1
All unit tests passing.

### DIFF
--- a/src/Data/ProgrammesDb/Entity/Ancestry.php
+++ b/src/Data/ProgrammesDb/Entity/Ancestry.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\MemberOfGroupInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\PromotableInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\RelatedLinkContextInterface;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Timestampable\Traits\TimestampableEntity;
+
+/**
+ * @ORM\Entity(repositoryClass="BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ImageRepository")
+ */
+class Ancestry implements MemberOfGroupInterface, RelatedLinkContextInterface, PromotableInterface
+{
+    use TimestampableEntity;
+    use Traits\IsEmbargoedTrait;
+    use Traits\PartnerPidTrait;
+    use Traits\SynopsesTrait;
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id()
+     * @ORM\Column(type="integer", nullable=true, unique=true)
+     * @ORM\ManyToOne(targetEntity="CoreEntity")
+     * @ORM\JoinColumn(nullable=true)
+     */
+    private $coreEntityId;
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id()
+     * @ORM\Column(type="integer",  nullable=true, unique=true)
+     * @ORM\ManyToOne(targetEntity="CoreEntity")
+     * @ORM\JoinColumn(nullable=true)
+     */
+    private $ancestorId;
+
+    public function __construct(string $ancestorId, string $coreEntityId)
+    {
+        $this->ancestorId = $ancestorId;
+        $this->coreEntityId = $coreEntityId;
+    }
+
+    public function getAncestorId(): ?int
+    {
+        return $this->ancestorId;
+    }
+
+    public function getCoreEntityId(): string
+    {
+        return $this->coreEntityId;
+    }
+
+    public function setAncestorId(string $pid): void
+    {
+        $this->ancestorId = $pid;
+    }
+
+    public function setCoreEntityId(string $pid): void
+    {
+        $this->coreEntityId = $pid;
+    }
+
+}
--- a/src/Data/ProgrammesDb/Entity/Ancestry.php
+++ b/src/Data/ProgrammesDb/Entity/Ancestry.php
@@ -17,20 +17,18 @@ class Ancestry implements MemberOfGroupInterface, RelatedLinkContextInterface, P
     /**
      * @var int
      *
-     * @ORM\Id()
      * @ORM\Column(type="integer", unique=true)
      * @ORM\ManyToOne(targetEntity="CoreEntity")
-     * @ORM\JoinColumn(name: "core_Entity_id", referencedColumnName: "id")
+     * @ORM\JoinColumn(name:"core_entity_id", referencedColumnName:"id")
      */
     private $coreEntityId;
 
     /**
      * @var int
      *
-     * @ORM\Id()
      * @ORM\Column(type="integer", unique=true)
      * @ORM\ManyToOne(targetEntity="CoreEntity")
-     * @ORM\JoinColumn(name: "ancestor_id", referencedColumnName: "id")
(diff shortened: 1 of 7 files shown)


kmurphy3377 (approved):
LGTM! 🚀

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id()
+     * @ORM\Column(type="integer",  nullable=true, unique=true)
+     * @ORM\ManyToOne(targetEntity="CoreEntity")
+     * @ORM\JoinColumn(nullable=true)
+     */
+    private $ancestorId;
```
colin-campbell-bbc, line 39: See above, same thing may need done for ancestor. To separate which columns in CoreEntity you are referencing, you may need to be more specific with `JoinColumn` https://www.doctrine-project.org/projects/doctrine-orm/en/2.15/reference/annotations-reference.html#annref_joincolumn

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @ORM\Entity(repositoryClass="BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ImageRepository")
+ */
+class Ancestry implements MemberOfGroupInterface, RelatedLinkContextInterface, PromotableInterface
+{
+    use TimestampableEntity;
+    use Traits\IsEmbargoedTrait;
+    use Traits\PartnerPidTrait;
+    use Traits\SynopsesTrait;
```
murkas, line 19: I think that we should be able to remove all of these traits as they apply to specific columns being available for the table.

src/Data/ProgrammesDb/Entity/Ancestry.php
kmurphy3377: There looks to be an example in Category if it helps any: https://github.com/bbc/programmes-pages-service/blob/c5dc28e27a2bdd31ac9db416b0eb15550ec47ed1/src/Data/ProgrammesDb/Entity/Category.php#L48-L52

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id()
+     * @ORM\Column(type="integer", nullable=true, unique=true)
+     * @ORM\ManyToOne(targetEntity="CoreEntity")
+     * @ORM\JoinColumn(nullable=true)
+     */
```
murkas: We will need to make all integer values non-nullable for the ancestry table:
```suggestion
    /**
     * @var int     *
     * @ORM\Id()
     * @ORM\Column(type="integer", nullable=false, unique=true)
     * @ORM\ManyToOne(targetEntity="CoreEntity")
     * @ORM\JoinColumn(nullable=false)
     */
```

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,56 @@
...
+    private $coreEntityId;
```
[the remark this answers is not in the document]
darceyh03: Oo good point, I'll change that now!

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @var int|null
+     *
+     * @ORM\Id()
+     * @ORM\Column(type="integer",  nullable=true, unique=true)
+     * @ORM\ManyToOne(targetEntity="CoreEntity")
+     * @ORM\JoinColumn(nullable=true)
+     */
```
murkas:
```suggestion
    /**
     * @var int
     *
     * @ORM\Id()
     * @ORM\Column(type="integer",  nullable=false, unique=true)
     * @ORM\ManyToOne(targetEntity="CoreEntity")
     * @ORM\JoinColumn(nullable=false)
     */
```

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,52 @@
...
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\MemberOfGroupInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\PromotableInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\RelatedLinkContextInterface;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Timestampable\Traits\TimestampableEntity;
+
+/**
+ * @ORM\Entity(repositoryClass="BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\ImageRepository")
```
ducampbe, line 12: This refers to `ImageRepository` - it makes me think we should have an AncestorRepository in the above location too (even if it has no meaningful methods)

src/Data/ProgrammesDb/Entity/Ancestry.php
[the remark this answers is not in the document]
darceyh03: I've had another look at this and I think since we're only using the id from core_entity that the type should remain as int. The ancestor collumn in core_entity currently only handles int values and since we're not planning to expand the information it holds in this project i think only taking the id for each entity will be fine!

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity;
+
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\MemberOfGroupInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\PromotableInterface;
+use BBC\ProgrammesPagesService\Data\ProgrammesDb\Entity\EntityInterfaces\RelatedLinkContextInterface;
+use Doctrine\ORM\Mapping as ORM;
+use Gedmo\Timestampable\Traits\TimestampableEntity;
```
murkas, line 9: These can also be cut down! 😄 
```suggestion
use Doctrine\ORM\Mapping as ORM;
```

src/Data/ProgrammesDb/Entity/Ancestry.php
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * @ORM\Entity(repositoryClass="BBC\ProgrammesPagesService\Data\ProgrammesDb\EntityRepository\AncestryRepository")
+ */
+class Ancestry implements MemberOfGroupInterface, RelatedLinkContextInterface, PromotableInterface
```
murkas, line 14: and in addition to the last one, this can also be simplified.
```suggestion
class Ancestry
```

colin-campbell-bbc (approved):

